PR #14, opened 2020-05-27
[SIEM]Dumps SIEM stuff

benskelker:
This just adds the SIEM docs folder as is. I'll need to update this when a couple of open 7.8 PRs are merged. Also, I still need to resolve the include from ml docs.


DonNateR:
> I'm fine with this, but also want to make sure @DonNateR is as well before this gets merged. I don't know if it will cause confusion for this to be in the `master` branch before 7.8 is released.

This should be ok. That's the normal workflow sooo

Issue #19, opened 2020-05-13
Document Timeline API and object schema


benskelker:
Add Timeline object schema to the documentation.

PR: https://github.com/elastic/security-docs/pull/50

cc @angorayc @XavierM @MikePaquette

PR #18, opened 2020-06-15
[Docs]Documents Timeline object schema

benskelker:
Moves the API and ref sections to first-level chapters.

~Adds Timeline API object schema.~

~[Timeline schema preview](https://security-docs_18.docs-preview.app.elstc.co/guide/en/security/master/timeline-object-schema.html)~

~All other changes are just for restructuring the new Elastic Security docs ToC, and they do not need to be reviewed.~

PR #50, opened 2020-07-05
[Docs]Adds timeline schema and API

benskelker:
Adds Timeline schema and API documentation.

[Timeline schema preview](https://security-docs_50.docs-preview.app.elstc.co/guide/en/security/master/timeline-object-schema.html)
[Timeline API overview preview](https://security-docs_50.docs-preview.app.elstc.co/guide/en/security/master/timeline-api-overview.html)
[Timeline create endpoint](https://security-docs_50.docs-preview.app.elstc.co/guide/en/security/master/timeline-api-create.html)
[Timeline import endpoint](https://security-docs_50.docs-preview.app.elstc.co/guide/en/security/master/timeline-api-import.html)

Supersedes PR: https://github.com/elastic/security-docs/pull/18, which only makes structural changes to ToC.

Resolves: https://github.com/elastic/security-docs/issues/19

### DIFF
--- a/docs/siem/timeline/api/timeline-api-create.asciidoc
+++ b/docs/siem/timeline/api/timeline-api-create.asciidoc
@@ -32,7 +32,7 @@ defines the search criteria and time range. The only required field is `title`.
 When you are creating a timeline template, provide these fields to so you can
 easily <<timeline-api-import, import template>> updates:
 
-* `templateTimelineId`: Unique identifier (for timeline templates only)
+* `templateTimelineId`: Unique identifier (UUID, for timeline templates only)
 * `templateTimelineVersion`: Template version number (for timeline templates
 only)
 
@@ -194,7 +194,7 @@ POST api/timeline
       "start": 1593832919000
     },
     "description": "Template for investigating host events",
-    "templateTimelineId": "HostSysProcess", <2>
+    "templateTimelineId": "6acb2c90-a01c-11ea-8e47-5dc21077d10c", <2>
     "templateTimelineVersion": 1, <3>
     "timelineType": "template", <4>
     "title": "Host event template"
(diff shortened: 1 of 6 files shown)


docs/siem/timeline/api/timeline-api-create.asciidoc
```diff
@@ -0,0 +1,360 @@
...
+        }
+      }
+    ],
+    "dateRange": {
+      "end": 1594005719000,
+      "start": 1593832919000
+    },
+    "description": "Template for investigating host events",
+    "timelineType": "template", <2>
+    "title": "Host event template"
```
angorayc, line 190: I think it would be better to also let users know that they call also provide a uuid as `templateTimelineId`, by doing this they can easily update the template by re-importing the template without looking up what templateTimelineId is assigned (as if they don't provide one, we'll assign one for them)

Also I would recommend to put `templateTimelineVersion` as `1` when users creating them, so they would aware that if they want to update the template by re-importing the file, they have to bump the version to any positive integer that's greater than that.

docs/siem/timeline/api/timeline-api-create.asciidoc
```diff
@@ -0,0 +1,462 @@
...
+          "value": "windows"
+        }
+      }
+    ],
+    "dateRange": {
+      "end": 1594005719000,
+      "start": 1593832919000
+    },
+    "description": "Template for investigating host events",
+    "templateTimelineId": "HostSysProcess", <2>
```
angorayc, line 197: Cool this would be useful! but we usually use uuid `e.g.: 6acb2c90-a01c-11ea-8e47-5dc21077d10c` for this field, so I think it would be better to align it.

PR #125, opened 2020-08-16
[Doc]Add Security requirements

benskelker:
Adds all requirements to a single location.

[Preview](https://security-docs_125.docs-preview.app.elstc.co/guide/en/security/master/sec-requirements.html)


benskelker:
@DonNateR @jmikell821 
The last commit adds the ingest page. It only really describes using beats to ingest data. I don't know if you want to build on it for the ingest manager, leave it as it is, or remove it - up to you.

docs/getting-started/detections-req.asciidoc
```diff
@@ -0,0 +1,124 @@
...
+[discrete]
+[[enable-detections-ui]]
+== Enable Detections
+
+To enable the <<detection-engine-overview, Detections feature>>, a user with
+these privileges must visit (click on) the *Detections* page:
+
+* The `manage` cluster privilege
+* {kib} space `All` privileges for the `Security` feature (see
+{kibana-ref}/xpack-spaces.html#spaces-control-user-access[Feature access based on user privileges]).
```
spong, line 47: We may want to mention adding `Saved Objects Management` here (or for users to initiate/create rules) just in case they're using this user for more than just `Enable Detections`, as these steps will allow them to 'unlock' the detections page, but the user will still see an error when creating a rule.

PR #129, opened 2020-08-17
Issue #63: Add Endpoint installation instructions

DonNateR:
Fixes issue #63


docs/getting-started/install-endpoint.asciidoc
```diff
@@ -0,0 +1,72 @@
...
+[float]
+== Add Elastic Security integration
+
+1. In Kibana, select **Security** > **Administration**. If this is not your first time using Elastic Security, select **Ingest Manager** > **Integrations** and search for **Elastic Endpoint Security**.
+2. On the Administration page of the security app or the Elastic Endpoint Security integration page in Ingest Manager, select **Add Endpoint Security**. The integration configuration page appears.
++
+[role="screenshot"]
+image::images/install-endpoint/security-integration.png[]
++
+3. Select a configuration for the Elastic Agent. You can use either the **Default config**, which works on most hosts, or adds security integration to a custom or existing configuration. For more details on Elastic Agent configuration settings, see {ingest-guide}/elastic-agent-configuration.html[Configuration settings].
```
EricDavisX, line 30: I'm not sure the context around 'which works on most hosts' ?  In the context of Endpoint Security, the Security Integration would be new and work on any host from the Default config.  I think we could remove that or clarify the intent of the message... if its about System Integration not working in some situation (System is the only Integration on by Default in the default config) we can call out concerns.  maybe better just to ignore them here tho, and focus on Endpoint?

docs/getting-started/install-endpoint.asciidoc
```diff
@@ -0,0 +1,72 @@
...
+3. After the download is complete, select an agent configuration. The selected configurations should include both **System** and **Elastic Endpoint Security** and any other integrations you chose to include when configuring an integration.
++
+[role="screenshot"]
+image::images/install-endpoint/endpoint-configuration.png[]
++
+4. After the Elastic Agent is installed on your host machine, open a command-line interface, and navigate to your Agent's directory. Copy the commands from Ingest Manager for your OS to enroll and run the Agent.
+
+=== Standalone mode
+
+If you want to manage the Elastic Agent without using Fleet to manage your hosts, you can enroll the Agent on your host in Standalone mode.
```
EricDavisX, line 61: I think we might want to link to the docs for this on the Ingest side instead of repeating it. This content below is specific to one deploy artifact group (.tar.gz for Linux or mac) and isn't correct for Windows, or .deb/.rpm (these 2 are the preferred method of installing on Linux, too) - we put significant effort in to get those good on Ingest side. :) 

Tho, for any reason, if we do want to keep *some specific instructions here... we probably should pick if we want it to be Windows specific maybe (as the most popular likely host type) or whichever OS and make sure the instructions are called out as being specific to that artifact type (.zip, .tar.gz, or .deb/.rpm).  And then we can link to the Ingest docs.  @dedemorton maybe can help confirm the desired doc page to link to

docs/getting-started/install-endpoint.asciidoc
```diff
@@ -0,0 +1,72 @@
...
+
+=== Enroll with Fleet
+
+1. Go to Ingest Manager. Select **Overview** > **Add agent**.
++
+[role="screenshot"]
+image::images/install-endpoint/add-agent.png[]
++
+2. In the Add agent pane of the Configurations section, download the Elastic Agent on your host's machine.
+3. After the download is complete, select an agent configuration. The selected configurations should include both **System** and **Elastic Endpoint Security** and any other integrations you chose to include when configuring an integration.
```
EricDavisX, line 52: I don't see a need to call out that System "should" be there?  It may or may not be, but you could cite that it is there in the default Ingest Manager configuration, and helps sending general monitoring and logs info on the host, separate and different info from the Endpoint, hence how it is helpful (something like that?).

docs/getting-started/install-endpoint.asciidoc
```diff
@@ -0,0 +1,72 @@
...
++
+[role="screenshot"]
+image::images/install-endpoint/endpoint-configuration.png[]
++
+4. After the Elastic Agent is installed on your host machine, open a command-line interface, and navigate to your Agent's directory. Copy the commands from Ingest Manager for your OS to enroll and run the Agent.
```
EricDavisX, line 57: this looks nice to me, glad we aren't trying to repeat steps (see note below for Stand-alone mode, maybe we shouldn't cite any specific 'command' steps there either...  and instead here, and there, we can link to the Ingest Manager docs page for installing Agent.  Thoughts?

EricDavisX: Having said that I see nothing bad about the above text, its great - I just think a link would be great too.  :)

docs/getting-started/install-endpoint.asciidoc
```diff
@@ -0,0 +1,72 @@
...
+1. In the Add agent pane, select the **Standalone mode** tab.
+2. Download the Elastic Agent on your host's machine and select an agent configuration.
+3. In the installed Agent's directory, create a file named `elastic-agent.yml`. Copy your configuration into the file from Step 3 of the Add agent pane. Remember to modify `ES_USERNAME` and `ES_PASSWORD` in the `outputs` section with your Elastic username and password.
+4. Save the file and start the Agent using `./elastic-agent run` command.
+5. Once the Agent begins to send data, verify that the data is viewable in Kibana by going to **Ingest Manager** > **Datasets**.
+
+[role="screenshot"]
+image::images/install-endpoint/datasets.png[]
+
+After you have enrolled the Elastic Agent on your host, select **Continue**. The host now appears on the Hosts view page inside the Elastic Security app.
```
EricDavisX, line 72: This isn't true for the stand-alone case, it feels a little out of place to have it here, maybe we could move it to the Fleet mode at the end, before the stand-alone case?

docs/getting-started/install-endpoint.asciidoc
DonNateR: @EricDavisX: Decided to add the OS commands underneath the image, like they appear in the the Ingest Manager docs. Let me know if that works.

EricDavisX: what ever you think is best - sounds good to me.

docs/getting-started/install-endpoint.asciidoc
```diff
@@ -0,0 +1,87 @@
...
+
+Like other Elastic integrations, Elastic Endpoint Security can be integrated into the Elastic Agent through {ingest-guide}/ingest-management-overview.html[Ingest Manager]. Upon configuration, the integration allows the Elastic Agent to monitor for events on your host and send data to the Elastic Security app.
+
+NOTE: Configuring the Endpoint Integration on the Elastic Agent requires that the user have permission to use Ingest Manager in Kibana.
+
+[discrete]
+[[security-before-you-begin]]
+== Before you begin
+
+If you're using the Elastic Agent on macOS Mojave (10.14) or later, ensure that you have enabled <<sensor-full-disk-access,Full Disk Access>>. Lastly, review the Elastic Security system requirements.
```
charlie-pichette, line 20: This takes you to the esensor steps not the Elastic Endpoint steps.  Enabling esensor will not work for the elastic-endpoint.

docs/getting-started/install-endpoint.asciidoc
```diff
@@ -0,0 +1,87 @@
...
+
+[discrete]
+[[add-security-integration]]
+== Add Elastic Security integration
+
+1. In Kibana, select **Security** > **Administration**. If this is not your first time using Elastic Security, select **Ingest Manager** > **Integrations** and search for "Elastic Endpoint Security".
+2. On the Administration page of the security app or the Elastic Endpoint Security integration page in Ingest Manager, select **Add Endpoint Security**. The integration configuration page appears.
++
+[role="screenshot"]
+image::images/install-endpoint/security-integration.png[]
```
charlie-pichette, line 30: This image seems to be more appropriate for step 1 since it displays the integration.  This image shows neither the add button nor the configuration page.  It is not until after you click on the Elastic Endpoint Security panel and then click the 'Add Elastic Endpoint Security' button that the integration configuration page appears.

docs/getting-started/install-endpoint.asciidoc
```diff
@@ -0,0 +1,87 @@
...
++
+3. Select a configuration for the Elastic Agent. You can use either the **Default config**, or adds security integration to a custom or existing configuration. For more details on Elastic Agent configuration settings, see {ingest-guide}/elastic-agent-configuration.html[Configuration settings].
+4. Configure the Elastic Endpoint Security integration with a name and options description. When done configuring, select **Save integration**. Kibana redirects you back to the administration section of the security app.
```
charlie-pichette, line 33: I think 'options description.' should be 'optional description.'

docs/getting-started/install-endpoint.asciidoc
DonNateR: Moved it to end of fleet section

dedemorton:
I don't have time to review this today because I have a bunch of release-related tasks to wrap up. But when the dust settles, let's talk about how we can provide a consistent set of instructions to users. I don't want to send users on an easter egg hunt, but I do think we should be using the same steps to describe the same procedure and that duplicating effort is not a good idea. Even if we decide to make the content available in multiple places, we should make sure we leverage each other's skills to make all the docs better. I think this will be easier when we put the ingest content into the new format where it's not buried in our big, text-only landing page.

docs/getting-started/install-endpoint.asciidoc
DonNateR: Are the steps different? Do you still have to enable full disk access for `elastic-endpoint`?

charlie-pichette: Yes, you still need to enable full disk access for elastic-endpoint.

DonNateR: Awesome. I'll update the Full Disk Access page in a separate PR.

PR #196, opened 2020-08-23
[Docs]Add banner to legacy SIEM docs

benskelker:


benskelker:
opening against the 7.9 branch only

PR #214, opened 2020-08-27
Add Kernel Endgame Inc instructions

DonNateR:
Not sure if I added this in the right place. Are these instructions specific to the legacy Endgame app @willyu-elastic and @ferullo?


ferullo:
The kernel extension needs to be approved for both products. However, as written these instructions apply to Elastic Endgame Security. I didn't check, but I trust this is already documented for Elastic Endgame.

PR #237, opened 2020-09-08
[Docs]Adds list item value parsing

benskelker:
Adds info on how to parse list item values and use custom value presentations.

[Preview](https://security-docs_237.docs-preview.app.elstc.co/guide/en/security/master/lists-api-create-container.html)


FrankHassanabad (approved):
LGTM! Thanks for the boiling down and simple example. Just the right amount of docs.

PR #272, opened 2020-10-07
[Docs]Adds EQL and threat-match rule types UI

benskelker:
NOTE: New updated PR: #336 

Adds new rule types to the UI documentation.

[Create rules preview](https://security-docs_272.docs-preview.app.elstc.co/guide/en/security/master/rules-ui-create.html)


docs/detections/rules-ui-create.asciidoc
```diff
@@ -29,13 +29,27 @@ generated for each value.
 For example, if the threshold `field` is `source.ip` and its `value` is `10`, an
 alert is generated for every source IP address that appears in at least 10 of
 the rule's search results.
+* *EQL*: Searches the defined indices and creates an alert when results match an
```
peluja1012, line 32: In the UI we decided to name this rule type "Event Correlation". I wonder if we should do that in this section also. @paulewing thoughts?

docs/detections/detections-ui-exceptions.asciidoc
```diff
@@ -61,15 +66,23 @@ To view, delete, or export existing lists:
...
 exception's criteria.
 
-IMPORTANT: When you select to close all alerts that meet the exception's
-criteria, all matching alerts are closed, *including* alerts generated by other
-rules.
+[IMPORTANT]
+==============
+Be careful when adding exceptions to EQL sequence rules. Exceptions are
+evaluated against every event in the sequence and when the exception matches a
+single event in the sequence, alerts are not generated. To exclude values from a
```
rylnd, line 79:
```suggestion
evaluated against every event in the sequence and when the exception matches any
event(s) in the sequence, alerts are not generated. To exclude values from a
```

Just trying to disambiguate the behavior here; the use of "matches a single event" is less accurate than "matches any event(s)" IMO

docs/detections/rules-ui-create.asciidoc
```diff
@@ -136,6 +150,55 @@ per rule execution:
 ** _Field_: Leave blank
 ** _Threshold_: `10`
 
+* To create a new EQL rule, select _Event Correlation_ and then:
```
rylnd, line 153: 👍 to @peluja1012 's [comment](https://github.com/elastic/security-docs/pull/272/files#r501724674) above: the inconsistency leads to weirdness like this.

docs/detections/rules-ui-create.asciidoc
```diff
@@ -136,6 +150,55 @@ per rule execution:
...
+.. Define which {es} indices the rule searches for alerts.
+.. Add an {ref}/eql-syntax.html[EQL statement] used to detect alerts.
++
+For example the following rule detects when `msxsl.exe` makes an outbound
+network connection:
+
+** _Index patterns_: `winlogbeat-*`
++
+Winlogbeat ship Windows events to {es-sec}.
+** _EQL query_: `sequence by process.entity_id [process where event.type in ("start", "process_started") and process.name == "msxsl.exe"] [network where event.type == "connection" and process.name == "msxsl.exe" and network.direction == "outgoing"]`
```
rylnd, line 163: Is there a way to make this a multiline code block? It's a bit hard to read as it is. Something like

```
sequence by process.entity_id
  [process 
    where event.type in ("start", "process_started") 
    and process.name == "msxsl.exe"]
  [network 
    where event.type == "connection" 
    and process.name == "msxsl.exe"
    and network.direction == "outgoing"]
```

Not sure if that's the "correct" EQL formatting, but you get the idea 🤷‍♂️

docs/detections/rules-ui-create.asciidoc
```diff
@@ -136,6 +150,55 @@ per rule execution:
...
+* To create a new EQL rule, select _Event Correlation_ and then:
+.. Define which {es} indices the rule searches for alerts.
+.. Add an {ref}/eql-syntax.html[EQL statement] used to detect alerts.
++
+For example the following rule detects when `msxsl.exe` makes an outbound
+network connection:
+
+** _Index patterns_: `winlogbeat-*`
++
+Winlogbeat ship Windows events to {es-sec}.
```
rylnd, line 162: Is there a convention for visually distinguishing an idea from its description? I _think_ that's what these two lines are, but I (and presumably other readers) would benefit from some visual categorization/reinforcement here.

rylnd (commented):
I can't speak to the Threat Match content, but EQL looks good! I'm going to defer to product for an approval, though.

docs/detections/rules-ui-create.asciidoc
DonNateR: @rylnd: I think so, let me explore this one.

docs/detections/rules-ui-create.asciidoc
DonNateR: @rylnd: Since I'm taking this over from Ben, I'm not sure which two ideas we're trying to distinguish. Are you hoping for a short flow of how the EQL query from winlogbeat moves from the process to the network?

rylnd: Hey @DonNateR ! Sorry, I was unclear there. I think the content is good, my question was centered around formatting:

Creating_detection_rules___Elastic_Security_Solution__master____Elastic

I was having trouble parsing that text and I _think_ it's because the rule parameter and its accompanying description aren't visually distinct. If that's standard formatting for something like this, though, I'm all for consistency 👍

peluja1012: @DonNateR aside from Ryland's comment around formatting, I think there is a typo in this line. I think it's supposed to say `Winlogbeat ships` as opposed to `Winlogbeat ship`.

docs/detections/detections-ui-exceptions.asciidoc
```diff
@@ -2,8 +2,14 @@
...
 == Rule exceptions and value lists
 
-To prevent the creation of unwanted alerts, you can add exceptions to detection
-rules. Exceptions contain the source event conditions that determine when
+To prevent the creation of unwanted alerts, you can add exceptions to these
+detection rule types:
+
+* Custom query
+* EQL
+* Threat match
```
peluja1012, line 10: @DonNateR @jmikell821 We recently renamed `Threat Match` to `Indicator Match`. I think you're already aware, but leaving this as a reminder. :)

PR #270, opened 2020-10-05
[Docs]Adds 7.10 detection rule updates to API docs

benskelker:
Updates the API docs with new rule types for 7.10.0.

[Create rule API preview](https://security-docs_270.docs-preview.app.elstc.co/guide/en/security/master/rules-api-create.html)
[Update rule API preview](https://security-docs_270.docs-preview.app.elstc.co/guide/en/security/master/rules-api-update.html)

PR #356, opened 2020-11-10
[7.10] [Docs]Adds 7.10 detection rule updates to API docs (#270)

jmikell821:
Backports the following commits to 7.10:
 - [Docs]Adds 7.10 detection rule updates to API docs (#270)


tylerjl:
Just to record this for posterity and future readers: this pull request was merged with the CLA check failing because the original author of the commit was a member of the Elastic organization at the _time the commit was authored_ but is not currently. Since that contribution was done while under membership of our organization, we (Infra) determined this was fine to merge in.

PR #372, opened 2020-11-13
Add updated detect index permissions

DonNateR:


spong (approved):
LGTM! 👍 Thanks for the quick fix here @DonNateR! 🙂

PR #383, opened 2020-11-30
Add known issue about value lists

DonNateR:
Remove redundant upgrade requirements section

### DIFF
--- a/docs/release-notes.asciidoc
+++ b/docs/release-notes.asciidoc
@@ -7,18 +7,6 @@
 :issue: https://github.com/elastic/kibana/issues/
 :pull: https://github.com/elastic/kibana/pull/
 
-[discrete]
-[[upgrade-notes]]
-== Post upgrade requirements
-
-After upgrading the {stack} to version 7.9.0 and 7.9.1 from a previous minor
-release (7.8.x, 7.7.x, and so on), you need to:
-
-* <<enable-detections-ui, Enable access to the Detections page>>. Previously
-activated detection rules continue to run after upgrading, and this is only
-required to enable the UI.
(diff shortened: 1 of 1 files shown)


jmikell821 (approved):
LGTM. You can merge once @peluja1012 approves the changes.

docs/release-notes.asciidoc
```diff
@@ -83,6 +71,8 @@ See ({pull}77327[#77327]) for details.
 [role="screenshot"]
 image::images/detection-rule-failure.png[]
 
+* When adding a rule exception, you can not select value lists with the `ip_range`. Any `ip_range` values will not appear in the dropdown even if you uploaded a value list with `ip_range` ({pull}79511[#79511]).
```
peluja1012, line 74: I think it would be more accurate if we update to say something like this:
> When adding a rule exception, you cannot select value lists of type `ip_range`. Lists of type `ip_range` will not appear in the Add Exception dropdown as possible values after selecting the `is in list` operator.

docs/release-notes.asciidoc
```diff
@@ -83,6 +71,8 @@ See ({pull}77327[#77327]) for details.
 [role="screenshot"]
 image::images/detection-rule-failure.png[]
 
+* When adding a rule exception, you cannot select value lists of type `ip_range`. Lists of type `ip_range` will not appear in the **Add Exception** dropdown as possible values after selecting the is in list operator. ({pull}79511[#79511]).
```
peluja1012, line 74: I would wrap "is in list" in tick marks ``

jmikell821: @DonNateR this has been sitting for 5 days. Please get this small change in and merge so we can close this out.

PR #426, opened 2021-01-04
DOCS: Clarification for using filters array

jguay:
using context.rule.filters is worth an example as it requires :
1- Understand that one should check the rule json representation to get an idea of which fields.
2- A bit of know-how on using mustache to iterate in an array and use a boolean condition


DonNateR:
Thanks for the contribution @jguay. Are you ok if I take over this PR?

jguay:
> Thanks for the contribution @jguay. Are you ok if I take over this PR?

@DonNateR Yes please

Issue #430, opened 2021-01-05
[DOCS] Detections Migration API


jmikell821:
[This topic](https://www.elastic.co/guide/en/security/current/post-upgrade-req.html) needs to be updated to include new reqs. 

Related issue: https://github.com/elastic/security-team/issues/204

DonNateR:
I believe this card will be taken care of by resolving issue #408, @jmikell821. Would you like me to assign this to myself?

jmikell821:
@DonNateR we have PR #435 open, but let me double check to make sure there's nothing else that needs to be added regarding this topic. We probably also need to move this to "System Reqs" - it hanging out at the bottom like that is kind of strange.

PR #435, opened 2021-01-06
Document Signals Migration APIs

rylnd:
Addresses #430. Outstanding Tasks:

- [ ] Decide where this page should live
- [ ] Decide where to reference this page
  - Perhaps this should just be mentioned in the 7.11 release notes? E.g. "Added Detection Alerts Migration API"


rylnd:
[Preview](https://security-docs_435.docs-preview.app.elstc.co/guide/en/security/master/signals-migration-api.html).

@jmikell821 I had intended for this all to be on one page, but I obviously don't totally understand asciidoc. Feedback/assistance would be appreciated!

jmikell821:
Hey @rylnd - thanks so much for doing this! There are no build errors so you did great. I'll review and will put everything on one page - then you can see my edits and we'll make sure everything looks good a final time before merging.

jmikell821:
@rylnd thanks again for your help with this! I added a `[float]` tag above the headings so they'd all be on the same page. I moved this to the API docs section, as you'll see in the preview. I added some grammatical edits, however, I do think this sentence is confusing and needs some clarity: 

>Finalize migrations
"As the finalization endpoint validates the migration’s status before completing and the finalization itself is idempotent, the endpoint can be used to poll for a given migration’s completion. During this step, for each successful migration, the original index’s alias is replaced by the migrated index’s alias."

Any suggestions on how we can make this more concise?

DonNateR (approved):

rylnd:
@jmikell821 thanks so much for the help here!



> Any suggestions on how we can make this more concise?

The intention here was to say: this endpoint is safe, you can call it repeatedly until it completes. Maybe:

> The finalization endpoint replaces original index's alias with the successfully migrated index's alias. As the endpoint is idempotent, it can safely be used to e.g. poll a given migration and, upon completion, finalize it.

DonNateR:
> @jmikell821 thanks so much for the help here!
> 
> > Any suggestions on how we can make this more concise?
> 
> The intention here was to say: this endpoint is safe, you can call it repeatedly until it completes. Maybe:
> 
> > The finalization endpoint replaces original index's alias with the successfully migrated index's alias. As the endpoint is idempotent, it can safely be used to e.g. poll a given migration and, upon completion, finalize it.

That sounds better to me @rylnd

docs/post-upgrade-req.asciidoc
```diff
@@ -1,11 +1,10 @@
...
-= Enable process analyzer after upgrade
+= Enable process analyzer after an upgrade
 
-To enable <<alerts-analyze-events, graphical representations of process relationships>>
-after upgrading to {stack} version 7.9.0 or 7.9.1 from a
-previous minor release (7.8.x, 7.7.x, and so on), you need to update
-`.siem-signals*` system index mappings.
+After upgrading from {stack} version 7.9.x to >= 7.10.x from a previous minor release (e.g., 7.8.x, etc.), you need to update `.siem-signals*` system index mappings to enable <<alerts-analyze-events, graphical representations of process relationships>>.
+
+NOTE: If you are upgrading from a minor release to {stack} version >=7.11.0, there is now a <<signals-migration-api>> that you can use instead of the manual process described below.
```
rylnd, line 7: Nit: there is no space between `>=` and `7.11.0` here, but there is one used in the previous sentence.

rylnd (commented):
LGTM! Thanks so much for the help @jmikell821 @DonNateR

PR #447, opened 2021-01-13
Correct source/destination.as field listings

iankoetter:
Change nested source/destination.as fields to the concrete sub fields as expected.
etc source.as -> source.as.number


ebeahan (approved):
Thanks, @iankoetter, for catching 👀 !

LGTM from the ECS team 👍

Issue #247, opened 2020-09-14
[DOCS}: Register Elastic Endpoint as anti-virus on Windows 


jmikell821:
## Description 
Add documentation for a user configuring Endpoint to register as the official AV on Windows

Issue: https://github.com/elastic/endpoint-dev/issues/7604

DonNateR:
Closing this since it is a duplicate of #407. Closing for now.

Issue #407, opened 2020-12-10
Register as AntiVirus [DOCS]


caitlinbetz:
## Description 
Users can toggle on the "Registered AntiVirus" option in their policy. 

## Acceptance Test Criteria

- Users can set Registered as AV via Policy
- Registered as AV is supported by 7.10+ Agents
- Windows only
- This config option is Basic license, NOT a tiered feature
- Turning this on will register Elastic Security as the AV solution in windows OS, and will disable windows defender

![image](https://user-images.githubusercontent.com/26076500/101825988-62e0c780-3afc-11eb-81ce-05422a9ac829.png)

DonNateR:
PR merged. Closing this issue now.

PR #451, opened 2021-01-15
Issue #407: Register Endpoint as Windows AV

DonNateR:
Also fixes issue #247.


docs/getting-started/install-endpoint.asciidoc
```diff
@@ -107,6 +107,19 @@ After you have installed the agent, malware prevention is automatically enabled
 [role="screenshot"]
 image::images/install-endpoint/malware-protection.png[]
 
+## Register as antivirus (Windows Only)
+
+If you download the Elastic Agent on Windows 10 or above, you can Elastic Security as your antivirus software. To enable Elastic Security as antivirus:
```
jmikell821, line 112:
```suggestion
If you download the Elastic Agent on Windows 10 or above, you can configure Elastic Security as your antivirus software by doing the following:
```

jmikell821: Consider changing `download` -> `downloaded` because they likely will have done so by the time they get to this step.

docs/getting-started/install-endpoint.asciidoc
```diff
@@ -107,6 +107,19 @@ After you have installed the agent, malware prevention is automatically enabled
...
+If you download the Elastic Agent on Windows 10 or above, you can Elastic Security as your antivirus software. To enable Elastic Security as antivirus: 
+
+1. Go to **Fleet** > **Agents**. A list of hosts with the Elastic Agent appears. 
+2. Select your desired host from the list.
+3. From the host's page, find your named **Endpoint** integration and select it. 
+4. On the Integration Edit page, look for the **Protections** section and find **Type: Register as antivirus**. Toggle this option to enable. 
++
+[role="screenshot"]
+image::images/register-as-antivirus.png[]
+5. Click **Save Integration**. Confirm that you would like to update your Elastic Agent. Once updated, Windows defender will be disabled on your local host in favor of Elastic Security.
```
jmikell821, line 121:
```suggestion
5. Click **Save Integration**. Confirm that you would like to update your Elastic Agent. Once updated, Windows Defender will be disabled on your local host in favor of Elastic Security.
```

jmikell821 (changes requested):
@DonNateR add the preview link in the comments please.

docs/getting-started/install-endpoint.asciidoc
```diff
@@ -107,6 +107,19 @@ After you have installed the agent, malware prevention is automatically enabled
 [role="screenshot"]
 image::images/install-endpoint/malware-protection.png[]
 
+## Register as antivirus (Windows Only)
+
+If you download the Elastic Agent on Windows 10 or above, you can Elastic Security as your antivirus software. To enable Elastic Security as antivirus: 
+
+1. Go to **Fleet** > **Agents**. A list of hosts with the Elastic Agent appears. 
+2. Select your desired host from the list.
+3. From the host's page, find your named **Endpoint** integration and select it.
```
kevinlog, line 116: This section makes it sound like the register as AV option will only apply to a selected host.  When you turn the option on, any Windows 10 Host with that particular Agent Policy will receive the "Register as AV" option.

I would drive the user towards choosing a Policy and informing then that any host using that Agent Policy would receive the Register as AV option

DonNateR: Kevin: The intention will be to move these steps into an overall Policy overview once this merged in. I think that's help set the context better.

kevinlog (approved):

Issue #429, opened 2021-01-05
[DOCS] Release Notes 7.11 


jmikell821:
Release notes for 7.11. We may have to add these manually. Let's loop in the `TEAM:Siem` and `Team:SecuritySolution` labels and see what we get. Let's generate at the latest on 1/21 so we can give the devs/leads time to review.

PR #458, opened 2021-01-20
Issue #429: 7.11 release notes

DonNateR:


kevinlog (approved):

kevinlog:
looks good from my perspective, we should also check with @peluja1012 since there are items in here about detections

jmikell821:
@DonNateR make sure to use transitive verbs (e.g., `adds`, `fixes`, etc.) for all entries to remain consistent with our other RNs. Some have it; others do not.

Issue #446, opened 2021-01-12
[DOCS] Document Detection Engine compatibility with Cold Tier storage and Searchable Snapshots


peluja1012:
Cold tier storage is also going GA in 7.11. Our users may end up running Detection Engine rules with data stored in this tier. We need to provide documentation that summarizes the Detection Engine's compatibility with Cold Tier storage. More details in this [ticket](https://github.com/elastic/security-team/issues/672).

The guidance we would like to document is below (more background [here](https://github.com/elastic/security-team/issues/672#issuecomment-756387528)):
- The security solution supports cold-tier data for the following Elasticsearch indices:

  - index patterns specified in securitySolution:defaultIndex
  - index patterns specified in the definitions of detection rules
  - index patterns specified in the data sources selector on the various security app pages

- The security solution does NOT support cold-tier data for the following Elasticsearch indices:
  - index patterns controlled by the Security solution, including signals and list indices
  - index patterns specified in Indicator Match rules as indicator index patterns

jmikell821:
Merged #487.

PR #487, opened 2021-02-01
[DOCS] Detection engine/cold tier data - 7.11

jmikell821:
Addresses #446. 

[Preview link](https://security-docs_487.docs-preview.app.elstc.co/guide/en/security/master/detection-engine-overview.html#cold-tier-detections).

### DIFF
--- a/docs/detections/detection-engine-intro.asciidoc
+++ b/docs/detections/detection-engine-intro.asciidoc
@@ -76,7 +76,7 @@ Cold tier is a {ref}/data-tiers.html[data tier] that holds time series data that
 {es-sec} does NOT support cold tier data for the following {es} indices:
 
 * Index patterns controlled by {es-sec}, including signals and list indices
-* index patterns specified in indicator match rules as indicator index patterns
+* Index patterns specified in indicator match rules as indicator index patterns
 
 If detection rules run with data stored in this tier, it likely will impact performance and speed.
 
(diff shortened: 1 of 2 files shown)


jmikell821:
@oatkiller @MikePaquette is there anything else I need to add here?

MikePaquette:
Thanks @jmikell821 this is looking good.

> If detection rules run with data stored in this tier, it likely will impact performance and speed.

This sentence could be clarified - are we providing more information the the cases that *are* supported, or are we saying what will happen if they run cold tier on *unsupported* indices?

jmikell821:
> Thanks @jmikell821 this is looking good.
> 
> > If detection rules run with data stored in this tier, it likely will impact performance and speed.
> 
> This sentence could be clarified - are we providing more information the the cases that _are_ supported, or are we saying what will happen if they run cold tier on _unsupported_ indices?

Hi @MikePaquette -- I'm not sure what else would happen beyond a sluggish performance if they ran cold tier on unsupported indices -- I suppose there could be a potential bug of the rule not executing but I don't want to assume. @oatkiller can you provide some more context please to Mike's question?

MikePaquette:
> If detection rules run with data stored in this tier, it likely will impact performance and speed.

@jmikell821 sorry my previous comment was not very clear.  I am asking that we tighten up that sentence because it is not clear to the reader what was meant by _run with data stored in this tier_.   The reader may wonder how does _run with data stored in this tier_ relate to the two lists you just gave me above?

Suggestion 1): delete the sentence.  (We've already said what's supported and what's not supported, do we need to elaborate?)
Suggestion 2): modify sentence as: "Using cold tier data for unsupported indices may result in detection rule timeouts and overall performance degradation.

MikePaquette:
One more suggestion @jmikell821, since this notice and topic is not strictly limited to the operation of the detection engine, can we either move it to or replicate it in the Elastic Security system requirements section?

image

oatkiller (approved):
Thanks!

Issue #412, opened 2020-12-16
[DOCS] Clarify that the example api requests we provided wouldn't work with Kibana Dev Tool Console


louisong:
## Description 
The [examples](https://www.elastic.co/guide/en/security/current/rules-api-create.html#_example_requests) we provided in our docs for Detection Rules API (this also applies to all other APIs in Elastic Security) seems to suggest that we can copy paste the examples provided in Kibana > Dev Tools and run them but they wouldn't work as this bypasses all the TaskManager/Alerting/Detections wrappers which could potentially messing things up.
@spong Thanks for confirming this!

In this example, i was trying to create a new rule using the "example 1" given in our api doc: https://www.elastic.co/guide/en/security/current/rules-api-create.html#_example_requests
via Kibana > Dev Tools and the api call went well (Chrome Dev Tools shows 200 response), Kibana > Dev Tools showed positive results like this:

```
{
  "_index" : "api",
  "_type" : "detection_engine",
  "_id" : "rules",
  "_version" : 5,
  "result" : "updated",
  "_shards" : {
    "total" : 2,
    "successful" : 2,
    "failed" : 0
  },
  "_seq_no" : 4,
  "_primary_term" : 1
}
```
The above result can give users a misleading impression that the rule was created successfully but in reality, this rule wasn't created at all. 

## Acceptance Test Criteria
As a user who is referencing Elastic Security APIs docs, he/she should be clearly aware that they must not copy/paste the [examples](https://www.elastic.co/guide/en/security/current/rules-api-create.html#_example_requests) via Kibana > Dev Tools console. 
To avoid confusion, we need to specifically call this out in our [Elastic Security API docs](https://www.elastic.co/guide/en/security/current/security-apis.html).

louisong:
@spong Would be good if you could help review this, thanks in adv!

PR #489, opened 2021-02-04
Issue #412 - Add warning about Kibana Dev Console

DonNateR:
Fixes issue #412

### DIFF
--- a/docs/siem-apis.asciidoc
+++ b/docs/siem-apis.asciidoc
@@ -2,6 +2,7 @@
 [[security-apis]]
 = Elastic Security APIs
 
+
 You can use these APIs to interface with {es-sec} features:
 
 NOTE: The {kib} Console supports only Elasticsearch APIs. You cannot interact with the {kib} APIs with the Console and must use `curl` or another HTTP tool instead. For more information, refer to https://www.elastic.co/guide/en/kibana/current/console-kibana.html[Console].
--- a/docs/siem-apis.asciidoc
+++ b/docs/siem-apis.asciidoc
@@ -2,7 +2,7 @@
 [[security-apis]]
 = Elastic Security APIs
 
-WARNING: Any request or response examples created in {kib} Dev Tool Console cannot be used to create an API request.
+WARNING: Do not use the {kib} Dev Tool Console to make any {es-sec} API requests. 
 
 You can use these APIs to interface with {es-sec} features:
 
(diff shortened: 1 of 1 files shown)


docs/siem-apis.asciidoc
```diff
@@ -2,6 +2,8 @@
 [[security-apis]]
 = Elastic Security APIs
 
+WARNING: Any request or response examples created in {kib} Dev Tool Console cannot be used to create an API request.
```
jmikell821, line 5: @DonNateR I don't think this is what we're trying to say. You _can_ make API requests via the dev tool console, you just can't copy and paste the examples; you have to type them in manually, I think. @louisong @spong is this correct?

DonNateR: My impression from reading the issue is that the call itself wouldn't work, even if the responses in the Dev Console were valid.

jmikell821: Gotcha - thanks. Slight suggestion.

```suggestion
WARNING: Do not use the {kib} Dev Tool Console to make any {es-sec} API requests. 
```

spong: Yeah specifically here, Kibana API requests cannot be made via the dev tools. This is the message used on the kibana docs here: https://www.elastic.co/guide/en/kibana/7.10/using-api.html. Perhaps we should just use that message as well?

image

spong (approved):
LGTM! Thanks @DonNateR! 🙂

docs/siem-apis.asciidoc
```diff
@@ -4,6 +4,8 @@
 
 You can use these APIs to interface with {es-sec} features:
 
+NOTE: The {kib} Console supports only Elasticsearch APIs. You are unable to interact with the {kib} APIs with the Console and must use `curl` or another HTTP tool instead. For more information, refer to https://www.elastic.co/guide/en/kibana/current/console-kibana.html[Console].
```
jmikell821, line 7: @DonNateR slight change for conciseness, then OK to merge. 

```suggestion
NOTE: The {kib} Console supports only Elasticsearch APIs. You cannot interact with the {kib} APIs with the Console and must use `curl` or another HTTP tool instead. For more information, refer to https://www.elastic.co/guide/en/kibana/current/console-kibana.html[Console].
```

jmikell821 (approved):
OK to merge.

Issue #488, opened 2021-02-02
[DOCS] Add additional case in Troubleshooting Missing Alerts  


MikePaquette:
## Description 

Add new material in https://www.elastic.co/guide/en/security/current/alerts-ui-monitor.html#troubleshoot-signals 

There would now be two cases to document  
1) when a rule fails to run close to its scheduled time  and you see gaps (existing) and 
2) Even if your rule runs at its scheduled time, there might still be missing alerts if your ingestion pipeline delay is greater than your rule interval + additional look-back time.

Please see slides 40-47 for a [slide-presentation](https://docs.google.com/presentation/d/1V71QdHdYzCEbKkFeAsz9P9bohA4absQ0ulnJgHdPsvU/edit?ts=600f618a#slide=id.gb94973f270_0_142) of what we want to add here for the documentation.

## Acceptance Test Criteria

A user who suspects they are missing alerts will find this section, read it, and be able to take steps to avoid missing further alerts.

## Notes

- Add the **"Team:Docs"** label to new issues. 
- Be sure to add the version number label. 
- Be sure to add any necessary screenshots for clarity. 
- Include any conditions or caveats that may affect customers.

jmikell821:
Merged #507.

PR #507, opened 2021-02-09
[DOCS] Troubleshoot missing alerts (7.11)

jmikell821:
Addresses #488.


docs/detections/rules-ui-monitor.asciidoc
```diff
@@ -30,3 +30,10 @@ If you see gaps for numerous rules:
 and then reactivating them in small batches at staggered intervals. This
 ensures {kib} does not attempt to run all the rules at the same time.
 * Consider adding another {kib} instance to your environment.
+
+Even if your rule runs at its scheduled time, there might still be missing alerts if your ingestion pipeline delay is greater than your rule interval + additional look-back time. As such, it is recommended that you use caution when creating custom rules with interval + additional look-back time values smaller than 6 minutes.
+
+You can also eliminate ingestion pipeline delay by specifying the `Timestamp override` field value to `event.digested` in <<rule-ui-advanced-params, advanced settings>> during rule creation. The detection engine uses the value from the `event.digested` field as the timestamp of when to execute the rule. For example, if an event occurred at 10:01 and was ingested in {es} at 10:10, the rule you created to detect that event would execute at 10:10 and generate an alert.
```
MikePaquette, line 36: Clarifications:
> You can also eliminate ingestion pipeline delay by specifying the `Timestamp override` field value to `event.digested` in <<rule-ui-advanced-params, advanced settings>> during rule creation. 

You can **eliminate the risk of missed alerts due to** ingestion pipeline delay by specifying the `Timestamp override` field value to **`event.ingested`** in <<rule-ui-advanced-params, advanced settings>> during rule creation **or editing.** 

> The detection engine uses the value from the `event.digested` field as the timestamp of when to execute the rule. For example, if an event occurred at 10:01 and was ingested in {es} at 10:10, the rule you created to detect that event would execute at 10:10 and generate an alert. 

The detection engine uses the value from the **`event.ingested`** field as the timestamp **when executing** the rule. For example, if an event occurred at 10:01, **and due to a 9-minute ingestion pipeline delay, was ingested into {es} at 10:10, a rule created to detect that event, with interval + additional look-back time of say 6 minutes, even if executing at 10:12, 11 minutes after the event occurred, would still detect the event, since the `event.ingested` timestamp was only two minutes earlier, well within the 6 minute rule query.**

MikePaquette (changes requested):
Thanks @jmikell821 I've requested a few clarifications below.

jmikell821:
Hi @MikePaquette one final glance please before I merge this. [Preview link](https://security-docs_507.docs-preview.app.elstc.co/guide/en/security/master/alerts-ui-monitor.html). Thanks!

docs/detections/rules-ui-monitor.asciidoc
```diff
@@ -31,9 +31,13 @@ and then reactivating them in small batches at staggered intervals. This
...
 
-You can also eliminate ingestion pipeline delay by specifying the `Timestamp override` field value to `event.digested` in <<rule-ui-advanced-params, advanced settings>> during rule creation. The detection engine uses the value from the `event.digested` field as the timestamp of when to execute the rule. For example, if an event occurred at 10:01 and was ingested in {es} at 10:10, the rule you created to detect that event would execute at 10:10 and generate an alert. 
+In addition, use caution when creating custom rule schedules to ensure that the specified interval + additional look-back time is greater than your deployment's ingestion pipeline delay.
+
+You can eliminate the risk of missed alerts due to ingestion pipeline delay by specifying the `Timestamp override` field value to `event.digested` in <<rule-ui-advanced-params, advanced settings>> during rule creation or editing. The detection engine uses the value from the `event.digested` field as the timestamp when executing the rule.
```
MikePaquette, line 38: `event.digested` should be `event.ingested` 2x

Otherwise LGTM, thanks!

MikePaquette (changes requested):
Thanks @jmikell821 
I added one comment with two corrections please.

docs/detections/rules-ui-monitor.asciidoc
jmikell821: Ah, good catch - thank you!

MikePaquette: although `event.digested` seems like a cool name for future field :-)

MikePaquette (approved):

PR #522, opened 2021-02-19
[DOCS] Updates Create Rule API docs for Threshold and Indicator Match Rules 

spong:
This is a rough pass at updating the [Create Rule API docs](https://www.elastic.co/guide/en/security/current/rules-api-create.html#_request_body) for the recent changes surrounding `Threshold` and `Indicator Match` rules. @madirey and @rylnd if you could verify these changes and if any additional fields should be documented that would be greatly appreciated. 🙂 

Resolves: https://github.com/elastic/kibana/issues/91965

Note: The grouping of optional fields per rule type makes these docs difficult to maintain (for me at least 😅), as you need to cross reference each optional field from each type, plus I'd bargain our users might find it useful to be able to look at all fields per each rule type independently. Something to verify with users/discuss, but would be a nice add for maintaining these docs (outside of writing scripts to generate the groupings).

##### Threshold Rules:
For more details on `cardinality_field` and `cardinality_value` see: https://github.com/elastic/kibana/pull/90826

##### Indicator Match Rules:

For more details on `threat_filters` see: https://github.com/elastic/kibana/pull/91260

### DIFF
--- a/docs/detections/api/rules/rules-api-create.asciidoc
+++ b/docs/detections/api/rules/rules-api-create.asciidoc
@@ -160,11 +160,11 @@ whether an alert is generated.
 |threshold |Object a|Defines the field and threshold value for when alerts
 are generated, where:
 
-* `cardinality_field` (string, required): The field on which the cardinality is applied. If
-you specify an empty field (`""`), alerts are generated when the query returns
-at least the number of results specified in the `value` field.
-* `cardinality_value` (integer, required): The threshold value from which an alert is
-generated based on unique number of values of cardinality_field.
+* `cardinality` (Array of length 1): The field on which the cardinality is applied.
+* `cardinality.field` (string, required): The field on which to calculate and compare the
+cardinality.k
+* `cardinality.value` (integer, required): The threshold value from which an alert is
+generated based on unique number of values of `cardinality.field`.
 * `field` (string or string[], required): The field on which the threshold is applied. If
 you specify an empty array (`[]`), alerts are generated when the query returns
 at least the number of results specified in the `value` field.
--- a/docs/detections/api/rules/rules-api-create.asciidoc
+++ b/docs/detections/api/rules/rules-api-create.asciidoc
@@ -160,8 +160,13 @@ whether an alert is generated.
 |threshold |Object a|Defines the field and threshold value for when alerts
 are generated, where:
 
-* `field` (string, required): The field on which the threshold is applied. If
+* `cardinality_field` (string, required): The field on which the cardinality is applied. If
 you specify an empty field (`""`), alerts are generated when the query returns
+at least the number of results specified in the `value` field.
+* `cardinality_value` (integer, required): The threshold value from which an alert is
+generated based on unique number of values of cardinality_field.
+* `field` (string or string[], required): The field on which the threshold is applied. If
+you specify an empty array (`[]`), alerts are generated when the query returns
 at least the number of results specified in the `value` field.
 * `value` (integer, required): The threshold value from which an alert is
 generated.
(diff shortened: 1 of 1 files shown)


docs/detections/api/rules/rules-api-create.asciidoc
```diff
@@ -336,9 +341,12 @@ a detection rule exception (`detection`) or an endpoint exception (`endpoint`).
 |==============================================
 |Name |Type |Description
 
-|threat_filter |Object[]
+|threat_filters |Object[]
 |{ref}/query-filter-context.html[Query and filter context] array used to filter
 documents from the {es} index containing the threat values.
+
+|threat_indicator_path |String
+|Much like an ingest processor, users can use this field to define where their threat indicator can be found on their indicator documents.
```
FrankHassanabad, line 349: @spong , @rylnd looks like the wording for threat_indicator_path is:

```
Much like an ingest processor, users can use this field to define where their threat indicator can be found on their indicator documents.
```

in the docs and on the UI it is:

```
Specify the document path containing your threat indicator fields. Used for enrichment of indicator match alerts. Defaults to threat.indicator unless otherwise specified.
```

Do you want it to be more like the UI for consistency and clarity?

```
Specifies the document path containing your threat indicator fields. Used for enrichment of indicator match alerts. Defaults to threat.indicator unless otherwise specified.
```

rylnd: IMO more variations is better for something like this: if a user still has questions after reading these docs, they may gain a better understanding of the concept after reading the field description in the rule form, or finding relevant commits/comments in kibana if they're particularly tech-savvy.

Note also that [the default value, and the field description](https://github.com/elastic/kibana/pull/92081) both changed for 7.12. I think it'd be useful for 7.12 docs to mention that the default here is `threatintel.indicator`.

docs/detections/api/rules/rules-api-create.asciidoc
```diff
@@ -160,9 +160,14 @@ whether an alert is generated.
 |threshold |Object a|Defines the field and threshold value for when alerts
 are generated, where:
 
-* `field` (string, required): The field on which the threshold is applied. If
+* `cardinality_field` (string, required): The field on which the cardinality is applied. If
```
madirey, line 163: Thanks @spong! This is the current behavior, but the field should not be required, and `cardinality_value` will only be required if `cardinality_field` is supplied. Will be updating this in an upcoming PR. I can change this here when that goes in...

docs/detections/api/rules/rules-api-create.asciidoc
rylnd, line 349:
```suggestion
|Much like an ingest processor, users can use this field to define where their threat indicator can be found on their indicator documents. Defaults to `threatintel.indicator`.
```

rylnd (approved):
LGTM but it'd be nice if we mentioned the default value here, too.

jmikell821 (approved):

docs/detections/api/rules/rules-api-create.asciidoc
spong: @madirey are we good to update this now with the latest fixes that made it in?

madirey: @spong Yeah, the schema looks a bit different now. Happy to update on Monday, or if you'd rather, I pasted the new schema below.

image

spong: Yeah if you could update that'd be great. I think this is the last change needed then we can merge 👍

Issue #546, opened 2021-03-10
[DOCS] Add information in the 7.11 and 7.12 troubleshooting section about ML job upgrade


dontcallmesherryli:
## Description 

Updated ML Job's were delivered in `7.11` (https://github.com/elastic/kibana/pull/85065) that take advantage of newly populated fields in the latest beats/agent. Since new ML Rules were not shipped alongside these job updates, if users update their beats/agent without first enabling the latest jobs and creating rules for them (or cloning existing and changing the selected ML Job), underlying detected anomalies may not have a Detection Alert generated and users would need to rely on the ML App visualization.

## Acceptance Test Criteria

Add troubleshooting information about upgrading ML jobs: if you upgraded to Elastic 7.11 or 7.12, your ML jobs might not work. At earliest convenience upgrade your ML rules to the newest version.

## Notes

- Add the **"Team:Docs"** label to new issues. 
- Be sure to add the version number label. 
- Be sure to add any necessary screenshots for clarity. 
- Include any conditions or caveats that may affect customers.

spong:
@jmikell821 -- both @randomuserid and I can help here with the details, so please reach out when this is ready to be worked 🙂

randomuserid:
So the revised language would look like this;

The conditions for the workaround are any of these;

1. Running a mixed version fleet of Elastic beats, including newer versions such as 7.10 and / or 7.11
2. Running a population of Elastic endpoints
3. Ingesting data from other sources using recent ECS versions

Workaround:

1. Clone the existing ML alert rules
2. Reconfigure each clone to query the v2 version of the target ML job

Note: there may be some edge cases of duplicate alerts, at the cost of eliminating dropped alerts, this needs analysis.

rylnd:
@jmikell821 @lcawl I've [specified](https://github.com/elastic/kibana/pull/94393/files#diff-9f3c81018ca4fe3e87d9791151923ec0f4ae0e85b440ecbe4f88b98739fea9cbR37) `ml-job-compatibility.html` as the prospective doc link for this issue. Is it possible to reverse-engineer the documentation page from this link? If not, what is the recommendation here?

lcawl:
> If not, what is the recommendation here?

[...]

PR #553, opened 2021-03-15
[DOCS] Add section for troubleshooting ML compatibility

lcawl:
Fixes https://github.com/elastic/security-docs/issues/546

### Preview

https://security-docs_553.docs-preview.app.elstc.co/guide/en/security/7.12/alerts-ui-monitor.html#ml-job-compatibility

### DIFF
--- a/docs/detections/rules-ui-monitor.asciidoc
+++ b/docs/detections/rules-ui-monitor.asciidoc
@@ -52,7 +52,40 @@ that are populated by {beats} and {agent} integrations. In version 7.11, new
 which operate on newer ECS fields than the previous
 <<security-winlogbeat-jobs>> and <<security-auditbeat-jobs>> jobs. However, the
 <<prebuilt-rules,prebuilt rules>> were not updated to use the new {ml} jobs.
-Therefore, if you have upgraded the cluster to 7.11 or 7.12 and are using
-multiple versions of {beats} and {agent}, to ensure continued rule coverage you 
-must create new rules that use the new {ml} jobs and enable them to run
-alongside your existing rules.
+Therefore, if you have upgraded your cluster and all of your {beats} and {agent} 
+integrations to 7.11 or later, you must update your existing rules to use the
+new jobs. If are still using earlier versions of {beats} and {agent} 
+integrations, however, you must create new rules that use the new {ml} jobs and 
+enable them to run alongside your existing rules.
+
+In particular, clone the following prebuilt rules and edit them to work with the
+new {ml} jobs:
+
+* <<unusual-linux-network-port-activity>>: Use 
+`v2_linux_anomalous_network_port_activity_ecs` instead of 
+`linux_anomalous_network_port_activity_ecs`.
+* <<anomalous-process-for-a-linux-population>>: Use 
+`v2_linux_anomalous_process_all_hosts_ecs` instead of 
+`linux_anomalous_process_all_hosts_ecs`.
+* <<unusual-linux-username>>: Use `v2_linux_anomalous_user_name_ecs` instead of 
+`linux_anomalous_user_name_ecs`.
+* <<unusual-linux-process-calling-the-metadata-service>>: Use 
+`v2_linux_rare_metadata_process` instead of `linux_rare_metadata_process`.
+* <<unusual-linux-user-calling-the-metadata-service>>: Use 
+`v2_linux_rare_metadata_user` instead of `linux_rare_metadata_user`.
+* <<unusual-process-for-a-linux-host>>: Use `v2_rare_process_by_host_linux_ecs` instead of `rare_process_by_host_linux_ecs`.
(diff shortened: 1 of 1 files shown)


docs/detections/rules-ui-monitor.asciidoc
```diff
@@ -41,3 +41,51 @@ For example, if an event occurred at 10:01, and due to a 9-minute ingestion pipe
...
+
+The <<prebuilt-ml-jobs,prebuilt {ml} jobs>> have dependencies on data fields
+that are populated by {beats} and {agent} integrations. In version 7.11, new
+{ml} jobs (<<security-linux-jobs>> and <<security-windows-jobs>>) were provided,
+which operate on newer ECS fields than the previous
+<<security-winlogbeat-jobs>> and <<security-auditbeat-jobs>> jobs. However, the
+<<prebuilt-rules,prebuilt rules>> were not updated to use the new {ml} jobs.
+Therefore, if you have upgraded your cluster and all of your {beats} and {agent} 
+integrations to 7.11 or later, you must update your existing rules to use the
+new jobs. If are still using earlier versions of {beats} and {agent}
```
spong, line 57:
```suggestion
new jobs. If you are still using earlier versions of {beats} and {agent} 
```

docs/detections/rules-ui-monitor.asciidoc
```diff
@@ -41,3 +41,51 @@ For example, if an event occurred at 10:01, and due to a 9-minute ingestion pipe
...
+that are populated by {beats} and {agent} integrations. In version 7.11, new
+{ml} jobs (<<security-linux-jobs>> and <<security-windows-jobs>>) were provided,
+which operate on newer ECS fields than the previous
+<<security-winlogbeat-jobs>> and <<security-auditbeat-jobs>> jobs. However, the
+<<prebuilt-rules,prebuilt rules>> were not updated to use the new {ml} jobs.
+Therefore, if you have upgraded your cluster and all of your {beats} and {agent} 
+integrations to 7.11 or later, you must update your existing rules to use the
+new jobs. If are still using earlier versions of {beats} and {agent} 
+integrations, however, you must create new rules that use the new {ml} jobs and 
+enable them to run alongside your existing rules.
```
spong, line 59: Is this accurate @randomuserid? My understanding was pre 7.11 beats/agent+rules+jobs were compatible. It's just that if you update your beats/agents different fields will be populated for which the new jobs are necessary. So as a result of a beats/agent upgrade, existing jobs/rules are incompatible, so you would need to upgrade to the new jobs/rules(manually). 

In this case though, do the new jobs work with earlier (pre-7.11) versions of the beats?

randomuserid: The new jobs use fields that did not exist when the first batch were made in 2019. So the new jobs work with current events, and the old jobs work with older events, with some exceptions as fields change.

lcawl: So does this mean it would be helpful to clarify this paragraph with something like this?:

> Therefore if you have upgraded your cluster and all of your Beats and Agent integrations to 7.11 or later, you must stop the Security:Auditbeat and Security:Winlogbeat jobs and use the Security:Linux and Security:Windows jobs instead. In this case, you must update your existing rules to use these new jobs. If you are still using earlier versions of Beats and Agent integrations, however, you can use both the old and new jobs and just stop a subset of the old jobs to avoid duplication (e.g. see https://www.elastic.co/guide/en/security/master/prebuilt-ml-jobs.html#_footnotedef_1). In this case, you must create new rules that use the new jobs and enable them to run alongside your existing rules.

randomuserid: It's complex enough to be hard to explain. As a user, I like seeing sets of logical options side by side in order to evaluate which set of conditions match my local conditions - maybe something like this;

- If running a fleet of Beats older then version 7.10, continue using the first generation ML jobs and rules that have been in the Security app since version 7.5.
- If running a fleet of Beats of version 7.10 or higher, use the second generation "v2" ML jobs, and optionally make clones of the existing ML rules, connected to the v2 jobs, in order to alert on anomalies from the v2 jobs.
- If running a mixed fleet of old and new Beats, or running a fleet of Beats and Elastic Endpoint agents, follow both of the previous steps to enable and use both the first and second generation ML jobs and rules.
- If running a non-Elastic data shipper to gather ECS-compatible Windows events, use the second generation "v2" ML jobs, and optionally make clones of the existing ML rules, connected to the v2 jobs, in order to alert on anomalies from the v2 jobs.

lcawl: Thanks for the details!  I've added that list to the doc, though I changed "7.10 or higher" to "7.11 or higher", since I don't see the Security:Windows or Security:Linux modules in 7.10 (https://github.com/elastic/kibana/tree/7.10/x-pack/plugins/ml/server/models/data_recognizer/modules).

PR #557, opened 2021-03-16
[DOCS] Removes [beta] tag from Detections intro

jmikell821:


nastasha-solomon:
Checked diff tab and looks like beta tag was removed.

PR #627, opened 2021-04-12
Clarify large value list exceptions depend on rule type

lucabelluccini:
The exceptions that are `is in list`/`is not in list` do not show up for threshold and eql rules.

Please add links to Threshold rules and Event correlation rules.


yctercero (approved):
Thanks so much!

lucabelluccini:
FYI I do not have rights so I'll let you manage the integration/backporting. Thank you @yctercero

jmikell821:
I'll take it from here @lucabelluccini. Thanks.

PR #701, opened 2021-05-25
[DOCS] Admin page updates (7.13) 

### LINKED ISSUE #669: [DOCS] Updates to Admin page
Following enhancements for 7.13: 

### Endpoint list: alignment with Agent statuses: https://github.com/elastic/security-team/issues/788
- updated agent statuses, updated name changes: Previously, we had an "error" status under the Agent Status column. Agent does not have the "Error" status but has "Unhealthy" - so we wanted to ensure we reflected the proper agent statuses

AC:

- When a Fleet Agent says "Unhealthy" for status, the corresponding Endpoint in Endpoint list also says "Unhealthy"
- When a Fleet Agent says "Healthy" for status, the corresponding Endpoint in Endpoint list also says "Healthy"
- When a Fleet Agent says "Offline" for status, the corresponding Endpoint in Endpoint list also says "Offline"
- When a Fleet Agent says "Updating" for status, the corresponding Endpoint in Endpoint list also says "Updating"
- If we do not recognize a status, we should use "Unhealthy"
- All statuses in Fleet Agent should match up in Endpoint list

### Trusted Applications - Hash validation https://github.com/elastic/security-team/issues/589

- If a user enters an invalid MD5, SHA256, and SHA1, the user will see an error in the form under the input and they will not be able to submit the form

### Trusted Applications - improved error messaging https://github.com/elastic/security-team/issues/714
- User receives an error message if they enter in multiple hash, path, or signer values

### Trusted Applications - Improved (simple)search
- The user can search for a Trusted Apps entry through the UI via search bar. Search bar allows users to search name, description, or value (hash, path, signer)

Note: this is not a KQL search

### Trusted Applications - Field descriptions https://github.com/elastic/security-team/issues/892
- added descriptions to the "Field" value when adding/editing a Trusted Application
image

### Trusted Applications - Users can edit a TA entry https://github.com/elastic/security-team/issues/682
- Users can edit a Trusted Application that has been previously added (by clicking the Edit button on the card)
- Users can edit the name, OS, field, operator, value, and description

Note: The Policy section of the trusted application entry (as seen in the ticket above) is NOT being released until a later date. For now users can only edit the existing trusted application fields.

jmikell821:
Resolves #669.


nastasha-solomon (approved):
LGTM!

PR #723, opened 2021-05-26
Remove "Copy as curl" and "View in console" links from examples

dedemorton:
It doesn't make sense to include "Copy as curl" or "View in console" here because the commands are not cURL, and you can't run them from the Dev Tools console.

### DIFF
--- a/docs/getting-started/install-endpoint.asciidoc
+++ b/docs/getting-started/install-endpoint.asciidoc
@@ -196,7 +196,7 @@ Follow these instructions to uninstall an endpoint **ONLY** if uninstalling an {
 
 Windows
 
-[source,console]
+[source,shell]
 ----------------------------------
 cd %TEMP%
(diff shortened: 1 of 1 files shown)


jmikell821 (approved):
@dedemorton makes sense, but if there are `[source]` command guidelines we should be following, can you link me to them? 

Also these commands would have looked more UI friendly in a widget, but I'm not sure how to add one, so if you have that info please pass along as well. Thank you.

dedemorton:
The guidelines are here:    https://github.com/elastic/docs/blob/master/README.asciidoc#code-blocks

That doc doesn't cover all the different syntaxes supported by asciidoctor, so you might also want to reference the general asciidoctor docs for more info about code blocks.

I have a [PR open with the widget code](https://github.com/elastic/observability-docs/pull/718/files) that you can copy, but you'll also need to get a copy of code.asciidoc to include the styles and script code because the regular docs build doesn't have that info. It's a little hacky, but works.

Issue #798, opened 2021-07-08
[DOCS] Document the new sub-feature privs for Cases 


nastasha-solomon:
Admins can now give a role varying levels of access to the Cases feature. To do this, they just need to configure the sub-feature **Cases** privs for **Security** when setting up role privs. Here is what each level allows:

- **`All`**: Can create/read/update/delete cases.
- **`Read`**: Can only read cases. 
- **`None`**: Permission to view the Cases page from the UI and nav menus is entirely revoked. Users can't see the Case page or access/edit cases at all. 

<img width="765" alt="case-privs" src="https://user-images.githubusercontent.com/79124755/124993915-05214900-e013-11eb-9abe-7e780255b57b.png">

It's also important to note that, because Cases use connectors, users who want the full range of abilities that come with managing cases (so creating/reading/updating/deleting them) also need to set their **Action and Connector** privs to **All**.

<img width="731" alt="RBAC-cases-connectors" src="https://user-images.githubusercontent.com/79124755/125837996-dafe85ce-a96c-4338-b210-1af35ecd79ca.png">

Considering the above, there are several combos that need to be documented in the [Cases prerequisites](https://www.elastic.co/guide/en/security/master/case-permissions.html) topic:

| Action      | Privs | 
| ----------- | ----------- | 
| User can **_manage_** cases     | **`All`** for `Cases` and **All** for **Action and Connector**       |
| User can only **_view_** cases    | **`Read`** for `Cases`      | 
| User cannot view or update Cases     | **`None`** for `Cases`      |

The last bit of important information to note is that sub-level privs for Cases override feature-level privs in the Security solution. Using the example below, this role should have read privs within the security solution app, and should not be able to see the cases page in the UI.

<img width="755" alt="case-sub-privs" src="https://user-images.githubusercontent.com/79124755/125855176-3886aa9e-4c3d-4524-b9a5-5c489d6a7f4a.png">

**Test env details**
- URL: https://source.kb.europe-west1.gcp.cloud.es.io:9243/
- Username: elastic
- Password: Gl6QDRN6X4Klt4JNPOn4P1hX

Remaining items to investigate:
- What priv combo do users need to manage rules? Do they need the currently prescribed privs in the [Detections prerequisites and requirements](https://www.elastic.co/guide/en/security/master/detections-permissions-section.html#enable-detections-ui) topic _and_ **`All`** set for **Action and Connector** privs?
- What index and cluster privs are needed with the different combos.

[...]

PR #814, opened 2021-07-16
[DOCS] Document the new sub-feature privs for Cases

nastasha-solomon:
Addresses #798.

Preview [here](https://security-docs_814.docs-preview.app.elstc.co/guide/en/security/master/case-permissions.html).


docs/getting-started/cases-req.asciidoc
```diff
@@ -8,4 +8,36 @@ For more information, see
...
-https://www.elastic.co/subscriptions[appropriate license].
\ No newline at end of file
+https://www.elastic.co/subscriptions[appropriate license].
+
+You can create roles and define feature privileges at different levels to manage feature access in {kib}. {kib} privileges grant access to features within a specified {kib} space and full or partial access can be given.
+
+When setting up feature access to cases, make sure to configure the following:
+
+- Base privileges for the {es-sec} solution
+- Sub-feature privileges for the Cases feature
```
jonathan-buttner, line 18: Should we indicate that this step is optional? They really only need to configure the sub-feature privileges for the Cases feature if they want it to be different than the base privilege. That might be implied though.

docs/getting-started/cases-req.asciidoc
```diff
@@ -8,4 +8,36 @@ For more information, see
...
+- Privileges for actions and connectors
+
+NOTE: Privileges set at the sub-feature level will override privileges set at the base level for the {es-sec} solution. For example, if a user's role has `All` privileges for the {es-sec} solution and `None` selected for the Cases feature, they'll have access to all features within the {es-sec} app except for cases.
```
jonathan-buttner, line 21: This is very helpful. Should we also note that if the user does not customize the sub-feature privilege the user will be granted the same access to cases? For example if they have All for Security Solution and have not customized the cases sub-feature (the toggle is disabled), the user will have All for cases.

jonathan-buttner (approved):
Looks good, just left a few comments.

docs/getting-started/cases-req.asciidoc
nastasha-solomon: Good point, this is an important distinction to make. I'll also include a line at the start of the topic describing what the default Case priv settings are.

docs/getting-started/cases-req.asciidoc
nastasha-solomon: Agree that this is a good example to include. I'll likely insert it right after the line I add about the default state of case privs.

docs/getting-started/cases-req.asciidoc
```diff
@@ -8,4 +8,36 @@ For more information, see
...
+image::images/cases-privs.png[]
+
+More information about the minimum privileges needed to use cases is below:
+
+[discrete]
+[width="100%",options="header"]
+|==============================================
+
+| Action      | {kib} Privileges
+| Give full access to manage cases
```
jmikell821, line 33:
```suggestion
| Give full access to manage Cases
```
(let's check cap vs. lowercase)

docs/getting-started/cases-req.asciidoc
```diff
@@ -8,4 +8,36 @@ For more information, see
...
+a|
+* `All` for the for the {es-sec} solution
+* `All` for the Action and Connector feature
+
+NOTE: Roles without `All` Action and Connector feature privileges cannot create, add, delete, or modify case connectors.
```
jmikell821, line 38:
```suggestion
* `All` for the Actions and Connectors feature

NOTE: Roles without `All` Actions and Connectors feature privileges cannot create, add, delete, or modify case connectors.
```

jmikell821 (commented):
@nastasha-solomon just a few comments and reorg suggestions. We'll loop Joe P. into the review after my and Jonathan's feedback is merged in. Thanks!

cnasikas (approved):

docs/getting-started/cases-req.asciidoc
```diff
@@ -1,11 +1,44 @@
...
-To view cases, you need the {kib} space `Read` privilege for the `Security` feature. To create cases and add comments, you need the `All` {kib}
-space privilege for the `Security` feature.
+//To view cases, you need the {kib} space `Read` privilege for the `Security` feature. To create cases and add comments, you need the `All` {kib}
+//space privilege for the `Security` feature.
 
-For more information, see
+//For more information, see
+//{kibana-ref}/xpack-spaces.html#spaces-control-user-access[Feature access based on user privileges].
+
+You can create roles and define feature privileges at different levels to manage feature access in {kib}. {kib} privileges grant access to features within a specified {kib} space and full or partial access can be given. For more information, see
```
joepeeples, line 10: Missing comma, and possible revision to make more active.
```suggestion
You can create roles and define feature privileges at different levels to manage feature access in {kib}. {kib} privileges grant access to features within a specified {kib} space, and you can grant full or partial access. For more information, see
```

docs/getting-started/cases-req.asciidoc
```diff
@@ -1,11 +1,44 @@
...
 {kibana-ref}/xpack-spaces.html#spaces-control-user-access[Feature access based on user privileges].
 
 NOTE: To send cases to external systems, you need the
-https://www.elastic.co/subscriptions[appropriate license].
\ No newline at end of file
+https://www.elastic.co/subscriptions[appropriate license].
```
joepeeples, line 14: Can this provide a bit more info on the type of license, or maybe the specific privilege required (using wording from the table on the target page)? That might save the user a click, or at least help them navigate the large table.

Something like "...you need an [appropriate license](https://www.elastic.co/subscriptions) that allows detection rule external actions" (or whatever the specific priv is from the list).

joepeeples (commented):
A couple of comments, otherwise looks good!

docs/getting-started/cases-req.asciidoc
nastasha-solomon: Completely agree with your suggestion, but am going to table this for now since I think this is the start of a larger convo.

Issue #770, opened 2021-06-18
[DOCS] Event Filters block Endpoint events from being streamed, so some custom Detection rules may not fire


kevinlog:
## Description 

TODO for now, scenarios is described here: https://github.com/elastic/kibana/issues/102619

## Acceptance Test Criteria

List all the ATC of each action and its intended result. 
For example: As a user, when [action (e.g., viewing, clicking, selecting, etc.)] I should [insert the expected result]. 

If the enhancement requires a procedure, include the steps.  

## Notes

- Add all appropriate labels to the issue, especially the version number label. 
- Be sure to add any necessary screenshots, code text or console commands for clarity. 
- Include any conditions or caveats that may affect customers.

jmikell821:
Merged #837.

PR #837, opened 2021-07-27
[DOCS] Event filters (7.14)

### LINKED ISSUE #741: Event Filtering [7.14][DOCS]
## Description 
Epic: https://github.com/elastic/security-team/issues/881

Users have events that are high volume/noisy and not helpful or useful to them (for example, metricbeat process events, or the same process making a network connection over and over to the same IP/port on a busy network server). We want to give users the ability to filter out events that they no longer want stored. We could do this similar to the way we do exceptions; drop events that match a given rule. This will make a user's storage in ES last longer.

## License
Basic

## Acceptance Test Criteria

**Users can add event filters two ways:**

1. Administration: Users can go to the "Event filters" tab in the Admin section and "Add Endpoint Event filter". This path will present the user with a blank form that they can fill out. 
image

image

<br />

2. Hosts tab > Events (sub-tab). Users can add an event filter directly from an Event. The "Add Endpoint Event Filter" option will be available on events that are from the Endpoint.  Doing this will populate the form automatically with fields/values related to the event. (Note: text currently says "event exception", this will be updated)

image

image

image

jmikell821:
Resolves #741; addresses #770. 

Preview [here](https://security-docs_837.docs-preview.app.elstc.co/guide/en/security/master/event-filters.html).


docs/management/admin/event-filters.asciidoc
```diff
@@ -0,0 +1,61 @@
...
+.. Go to *Manage* -> *Event filters*.
+.. Click *Add Event Filter*.
++
+The *Add event filter* flyout opens.
++
+[role="screenshot"]
+image::images/event-filter.png[]
+. Enter a name for the event filter.
+. Depending on from which page you add the filter, either modify the pre-populated conditions that appear, or
+add new conditions that define when {elastic-sec} filters events. You can define multiple conditions with `OR` and `AND` relationships. You can also add nested conditions. In the image above, the event filter filters out events whose `event.category` field is `network`, and whose `process.executable` field is the same as the specified path.
```
kevinlog, line 31: You can only create Event Filters with `AND` relationships for now.  `OR` will always be disabled

kevinlog (approved):
I left a small comment to change, but after that LGTM!

caitlinbetz (approved):

nastasha-solomon (approved):
LGTM!

docs/management/admin/event-filters.asciidoc
```diff
@@ -0,0 +1,60 @@
...
+. To create an event filter via the Event filters page:
+.. Go to *Manage* -> *Event filters*.
+.. Click *Add Event Filter*.
++
+The *Add event filter* flyout opens.
++
+[role="screenshot"]
+image::images/event-filter.png[]
+. Enter a name for the event filter.
+. Depending on which page you add the filter, either modify the pre-populated conditions or add new conditions that define when {elastic-sec} filters events. You can define multiple conditions with `OR` and `AND` relationships. You can also add nested conditions. In the image above, the event filter filters out events whose `event.category` field is `network`, and whose `process.executable` field is the same as the specified path.
```
joepeeples, line 30:
```suggestion
. Depending on which page you add the filter, either modify the pre-populated conditions or add new conditions that define when {elastic-sec} filters events. You can define multiple conditions with `OR` and `AND` relationships. You can also add nested conditions. In the image above, the event filter excludes events whose `event.category` field is `network`, and whose `process.executable` field is the same as the specified path.
```

docs/management/admin/event-filters.asciidoc
```diff
@@ -0,0 +1,60 @@
...
+. Add a new comment that describes identifies the filter (optional).
+. Click *Add Endpoint Event Filter*. The new filter is added to the Event filters table.
+
+[[manage-event-filters]]
+[discrete]
+== View and manage event filters
+
+The Event filters table allows you to view and manage all endpoint event filters that have been added. To view the Event filters table, go to *Manage* -> *Event filters*. Event filters appear in reverse chronological order, with the most recently modified filter at the top. Each filter has its own "card," which displays details such as the filter name, operating system, date created, and the filter conditions.
```
joepeeples, line 38: Looking at the UI in Kibana, this seems more like an "Event Filter _page_" instead of a "table," because the info isn't really organized in a table format. Is possible to change this term?

Similarly, I would suggest _section_ instead of _card_ (minus the quotation marks), because if something is a card I would expect to be able to move, shuffle, or somehow interact with it, but these appear to be just static sections of the page.

docs/management/admin/event-filters.asciidoc
```diff
@@ -0,0 +1,60 @@
...
+. Click *Edit* on the appropriate event filter card.
+. Modify details or conditions as appropriate.
+. Click *Update event filter*. The newly modified filter appears at the top.
+
+To delete an event filter:
+
+. Click *Remove* on the event filter to delete.
+. On the dialog that appears, confirm that you are removing the correct event filter, then click *Remove event filter*.
++
+A message confirming the filter was removed appears.
```
joepeeples, line 60:
```suggestion
A message appears confirming the filter was removed.
```

joepeeples (commented):
A few comments and line edits.

docs/management/admin/event-filters.asciidoc
jmikell821: @joepeeples the design team is discussing what to call these UI elements (e.g. `table` vs. `list`, etc.) I am currently following what's in the trusted applications topic. I also previously asked them about the usage of `card` - which is correct.

PR #872, opened 2021-08-03
Link siem-apis.asciidoc to the new Fleet api docs page

EricDavisX:
This change is just to support navigability / searchability for users to find the Fleet APIs if they are also using the siem / security apis.

work discussed in https://github.com/elastic/observability-docs/issues/881

if you like it?  enjoy.


spong (approved):
LGTM! Thanks for adding this @EricDavisX! 🙂

Issue #1014, opened 2021-09-07
[DOCS] Add more context to indicator match rule restrictions


jmikell821:
Topic to update: [here](https://www.elastic.co/guide/en/security/current/detection-engine-overview.html#support-indicator-rules). 

Related issue comment [here](https://github.com/elastic/security-team/issues/1682#issuecomment-911495528). 

Let's expand the text on tightening up search queries to limit unwanted data. 

> Indicator match rules provide a powerful capability to search your security data; however, their queries can consume significant deployment resources. When creating an <<create-indicator-rule, indicator match rule>>, it is recommended to limit queries to a specific time range to avoid results that return more indicators than necessary for the desired rule coverage. For example, entering a query such as `@timestamp > "now-30d"` will return results from the past 30 days only. 

Hi @rylnd if there's anything else we need to add to this blurb here ^ please let us know. Also, please review for accuracy. Thank you!

nastasha-solomon:
Merged #1025.

PR #1025, opened 2021-09-10
[DOCS] Add more context to indicator match rule restrictions

nastasha-solomon:
Addresses #1014.

Preview [here](https://security-docs_1025.docs-preview.app.elstc.co/guide/en/security/master/detection-engine-overview.html#support-indicator-rules).


docs/detections/detection-engine-intro.asciidoc
```diff
@@ -82,7 +82,7 @@ Using cold tier data for unsupported indices may result in detection rule timeou
 [[support-indicator-rules]]
 == Limited support for indicator match rules
 
-Indicator match rules provide a powerful capability to search your security data; however, their queries can consume significant deployment resources. We recommend avoiding indicator index queries that return more indicators than necessary for the desired rule coverage. In addition, the following support restrictions are in place:
+Indicator match rules provide a powerful capability to search your security data; however, their queries can consume significant deployment resources. When creating an <<create-indicator-rule, indicator match rule>>, it is recommended to limit queries to a specific time range to avoid results that return more indicators than necessary for the desired rule coverage. For example, entering a query such as `@timestamp > "now-30d"` will return results from the past 30 days only.
```
ecezalp, line 85: > avoid results that return more indicators than necessary

I don't think that we want to `avoid results`, also `return more indicators than necessary` sounds a bit suboptimal as the problem is less about returning the indicators and more about querying too many _irrelevant_ indicators. We could rephrase along the lines of 

> When creating an <<create-indicator-rule, indicator match rule>>, it is recommended to reduce the number of queried indicators for performance benefits. Using an indicator index query such as `@timestamp > "now-30d"` ensures that only recent indicators (ingested in the past 30 days) would be used for threat matching.

MikePaquette: ++ for avoiding `avoid results` :-)

Our primary persona for this section is the detection engineer.  Her primary goal at work is to achieve effective detections of bad stuff, not to obtain performance benefits.  So we should empathize with this reader persona by highlighting "desired rule coverage" as the primary objective, rather than performance benefits.

Also, we need to make it clear that we're talking about the indicator index query, not the custom query in the rule.

I think this slight tweak works better:

When creating an <<create-indicator-rule, indicator match rule>>, it is recommended to limit the time range of the indicator index query to the minimum time period necessary for the desired rule coverage. For example, entering a query such as `@timestamp > "now-30d"` will include only indicators ingested during the past 30 days when the rule executes.  [OPTIONAL]  Without this time period limitation, the rule will include *all* indicators in your indicator indices, which may cause excessively long queries.

rylnd: One more note: It looks like you lost the segue to the subsequent list in these changes; the "the following support restrictions are in place:"

docs/detections/detection-engine-intro.asciidoc
```diff
@@ -82,7 +82,7 @@ Using cold tier data for unsupported indices may result in detection rule timeou
 [[support-indicator-rules]]
 == Limited support for indicator match rules
 
-Indicator match rules provide a powerful capability to search your security data; however, their queries can consume significant deployment resources. We recommend avoiding indicator index queries that return more indicators than necessary for the desired rule coverage. In addition, the following support restrictions are in place:
+Indicator match rules provide a powerful capability to search your security data; however, their queries can consume significant deployment resources. When creating an <<create-indicator-rule, indicator match rule>>, we recommend limiting the time range of the indicator index query to the minimum period necessary for the desired rule coverage. For example, entering a query such as `@timestamp > "now-30d"` will return indicators ingested during the past 30 days. Without this limitation, the rule will include all of the indicators in your indicator indices, which may cause excessively long queries.
```
jmikell821, line 85: @nastasha-solomon general question about this: when we say `"may cause excessively long queries"` are we referring to the number of results returned? I want to verify we are using `query` in the right context. Just wanted to double check.

MikePaquette:
Long queries refers to the length of time it takes them to complete.

MikePaquette (approved):
LGTM, thanks for making these improvements!

docs/detections/detection-engine-intro.asciidoc
nastasha-solomon: Yeah, I saw that and just need to commit my changes to add it back in. Thanks anyway for pointing it out @rylnd :)

rylnd (approved):
LGTM!

docs/detections/detection-engine-intro.asciidoc
nastasha-solomon: @jmikell821 based on @MikePaquette 's response, would the following revision work:
Without this limitation, the rule will include all of the indicators in your indicator indices, which may extend the time it takes for the indicator index query to complete.

nastasha-solomon: Inserting the following:
Without this limitation, the rule will include all of the indicators in your indicator indices, which may extend the time it takes for the indicator index query to complete.

PR #1171, opened 2021-10-20
[Docs] [Security Platform] adds warning for basic auth only

dhurley14:
Alerting docs added a warning to their docs to inform users that the create / update / enable api's only allow basic auth. Because we rely on the alerting api's we need to bubble up this same warning in our docs too.


jmikell821:
@elasticmachine run elasticsearch-ci/docs

PR #1249, opened 2021-11-24
[DOCS] Alerts page updates (7.16)

### LINKED ISSUE #1196: [DOCS] Alerts table updates (7.16)
Though this work was done by the threat hunting team, it's more of a D&R feature, hence the two labels. 

Topic: [View and filter detection alerts](https://www.elastic.co/guide/en/security/current/alerts-ui-manage.html#detection-view-and-filter-alerts)

Threat Hunting Feature Sync Notes [here](https://docs.google.com/document/d/19CQVJxs1V7Lg3EoYdhtIppmQcprpAQ_CGIuHoxSS-fc/edit?usp=sharing).

There have been some enhancements to the Alerts table: 

- Analyze event appears in the table, where available, provided they have entered the correct KQL query. 
**Note:** This feature is also available in `7.15.2`, to be documented in #1191. 

- Default inline actions have been limited to `filter in, filter out, add to timeline`, and ` expand`. 

- `Reason` field renamed and no longer a default field/table column

Users need to manually add the reason field to the Alerts table. To do this: 
   1. Click **Fields**. 
   2. Enter  `signal.reason` into the search bar. 
   3. Select the `signal.reason` field to add it to the Alerts table. 
   4. Click **Close** to return to the Alerts table. 
 

**Note:** In `v8.0`, the `reason` field will be renamed to `kibana.alert`. 
The **Expand** button is the only inline action available for the `reason` field. When selected, the complete reason for why the alert was generated displays in the event renderer view.

joepeeples:
Resolves #1196.

Previews:

- [Manage detection alerts](https://security-docs_1249.docs-preview.app.elstc.co/guide/en/security/master/alerts-ui-manage.html)

Note: As far as I can tell, the `Reason` field is still included by default in the data grid in 7.16 -- I confirmed this in two brand-new cloud deployments. Maybe this will be removed in 8.0 instead? If so, we can add steps on how to add the field in that version.


docs/detections/alerts-ui-manage.asciidoc
```diff
@@ -23,8 +23,13 @@ NOTE: KQL autocomplete for `.siem-signals-*` indices is available on the
...
-* Group alerts by specific parameters and view individual alert counts by using the *Stack by* drop-down in the Count table. For example, if you select `signal.rule.name`, the table displays the number of alerts generated by each rule.
+* Group alerts by specific parameters and view individual alert counts by using the *Stack by* drop-down in the *Count* table. For example, if you select `signal.rule.name`, the table displays the number of alerts generated by each rule.
+* Visualize alerts by specific parameters by using the *Stack by* drop-down in the *Trend* histogram. For example, if you select `signal.rule.name`, the histogram displays the total alert counts by rule name.
+* Hover over a value in the data grid to display the inline actions available, such as *Filter In*, *Filter Out*, and *Add to timeline investigation*. Click the expand icon to open a full context menu of options, including *Show top values* and *Copy to Clipboard*. The options available vary depending on the type of data.
```
nastasha-solomon, line 28: Just have one small nitpick/question: do we call the expand icon the **View details** button or is that a different button than the one that displays when you hover over a value in the data grid?

nastasha-solomon (approved):
Other than one question, LGTM!

docs/detections/alerts-ui-manage.asciidoc
joepeeples: @nastasha-solomon They're two different UI elements, with similar-looking icons. The `View details` button opens the flyout with all of the data from the selected alert, while the expand icon that appears upon hovering (which unfortunately doesn't have a name) opens a context menu of actions you can perform on the selected value.

joepeeples:
> looks good to me. Only thing I noticed is I am not sure how the first gif connects to the text around it. It shows the menu for an alert and then selecting View rule details, but I didn't find view rule details until later in the page. Still, it does show the interface the page is discussing so maybe this is all good.

@benironside Good point. I didn't mean to illustrate selecting `View rule details`; I just wanted to pop open the context menu and then generally gesture at the wide range of options available on that menu. But I can see how that would be interpreted as a deliberate selection. I'll re-capture the gif without that extra mouse movement at the end.

docs/detections/alerts-ui-manage.asciidoc
```diff
@@ -23,8 +23,13 @@ NOTE: KQL autocomplete for `.siem-signals-*` indices is available on the
...
 * View detection alerts generated by a specific rule. To do this, go to *Detect* -> *Rules*, then select a rule name in the All rules table. The *Rules detail page* displays a comprehensive view of the rule's details, and alert details are displayed in the Alerts table beneath the Trend histogram. Alternatively, you can go to *Detect* -> *Alerts* -> *Manage rules*, then click on a rule name in the All rules table to view the alerts associated with the rule.
-* Visualize alerts by specific parameters by using the *Stack by* drop-down in the Trend histogram. For example, if you select `signal.rule.name`, the histogram displays the total alert counts by rule name.
-* Group alerts by specific parameters and view individual alert counts by using the *Stack by* drop-down in the Count table. For example, if you select `signal.rule.name`, the table displays the number of alerts generated by each rule.
+* Group alerts by specific parameters and view individual alert counts by using the *Stack by* drop-down in the *Count* table. For example, if you select `signal.rule.name`, the table displays the number of alerts generated by each rule.
```
michaelolo24, line 26: Guessing the re-order was because of the reordering of the items in the view? Was just curious

michaelolo24 (approved):
Looks good, thanks @joepeeples ! LGTM 😄

docs/detections/alerts-ui-manage.asciidoc
joepeeples: Yep! Just matching the order of the items on the page, top to bottom, left to right.

docs/detections/alerts-ui-manage.asciidoc
jmikell821, line 28: We'll be having more discussions of icon vs. button likely at our next UX doc leads meeting. But I like to think of icon as a representation of something, but while you can click it, I think of "button" more of something in a menu, etc. that you can click. The ability to add inline images in the new docs system will make describing these graphic elements in the menus easier. 

Interesting tidbit from the Microsoft Style Guide about use of "[icon](https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/i/icon)". 
```suggestion
* Hover over a value in the data grid to display available inline actions, such as *Filter In*, *Filter Out*, and *Add to timeline investigation*. Click the expand button to open a full context menu of options, including *Show top values* and *Copy to Clipboard*. The options available vary depending on the type of data.
```

joepeeples: I agree with that distinction, and looking forward to more discussions and guidelines on word choices like this!

PR #1376, opened 2022-01-06
[DOCS] update list endpoints documentation

joeypoon:
Updating management endpoints lists API docs from old POST method to new GET method.

Preview [here](https://security-docs_1376.docs-preview.app.elstc.co/guide/en/security/master/list-endpoints-api.html).

### DIFF
--- a/docs/management/api/list-endpoints-api.asciidoc
+++ b/docs/management/api/list-endpoints-api.asciidoc
@@ -7,33 +7,23 @@ NOTE: Console supports Elasticsearch APIs only. Console doesn't allow interactio
 
 ==== Request URL
 
-`POST <kibana host>:<port>/api/endpoint/metadata`
+`GET <kibana host>:<port>/api/endpoint/metadata`
 
 
-==== Request body
+==== URL query parameters
 
-A JSON object with these fields:
+All parameters are optional:
 
 [width="100%",options="header"]
 |==============================================
-|Name |Type |Description |Required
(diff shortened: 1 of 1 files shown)


pzl:
@joeypoon how far back did the POST-->GET and pagination changes get merged? 

- Did it get backported to anything in `7.x`?
- Did we (ever) delete the POST route?
  - assuming we did, that occurred in 8.0, yes?
  
We need to keep the POST docs around for whatever versions it existed until deletion (8.0, if we did this correctly).

If the GET route was backported to 7.x, I'm fine _also_ adding that as a way to list endpoints (so, we need to document the existence of both simultaneously). So this diff should be almost all additions, explaining that both routes exist. Then perhaps a second PR that removes the POST info and doesn't get backported to 7.x.

If GET wasn't added until 8.0, and POST was simultaneously deleted for the same release, then this diff is OK, it just needs to be in 8.0 docs and beyond, no `7.x` backporting.

jmikell821:
@pzl I added the `7.n` versions to the PR, so that was my fault. Joey had just listed `8.0`.

joeypoon:
@pzl everything went in 8.0.0, including deletion of the POST.

thanks for adding the preview, @jmikell821, I wasn't sure how to do that :). should we remove the `7.x.x` labels?

pzl (approved):
This is :+1:  for 8.0. I removed the other labels

Issue #1416, opened 2022-01-18
[DOCS] `8.0-rc2` release notes  


jmikell821:
Include these in RNs for `8.0-rc2` (I'll set up the syntax in the docs) 

- Check out `main`
- Edit RN files as appropriate
- Merge to `main`, backport to `8.0`. 

PRs to include: 

## Known issues

- [ ] https://github.com/elastic/kibana/issues/120906
- [ ] https://github.com/elastic/kibana/issues/121390

PR #1425, opened 2022-01-20
[DOCS] 8.0-rc2 release notes

nastasha-solomon:
Addresses https://github.com/elastic/security-docs/issues/1416. 

Preview:
- [8.0.0-rc2 Release notes](https://security-docs_1425.docs-preview.app.elstc.co/guide/en/security/master/release-notes-header-8.0.0.html)
- [Updated index page for 8.0.0-rc2](https://security-docs_1425.docs-preview.app.elstc.co/guide/en/security/master/release-notes.html)


docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
...
+The role must have the following privileges:
+
+* *Cluster privileges*: The `manage` privilege
+* *Index privileges*: The `manage`, `write`,`read`, and `view_index_metadata` index privileges for the following system indices where `<kib-space>` is the {kib} space name:
+
+** `.siem-signals-<kib-space>`
+** `.lists-<kib-space>`
+** `.items-<kib-space>`
+** `.alerts-security.alerts-<kib space>`
+** `.internal.alerts-security.alerts-<kib space>-*`
```
joepeeples, line 27:
```suggestion
** `.alerts-security.alerts-<kib-space>`
** `.internal.alerts-security.alerts-<kib-space>-*`
```

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
...
+*Rules must be manually re-enabled after upgrading:*
+
+Rules are automatically disabled when upgrading to {stack} version 8.0.0 and must be manually re-enabled after completing the upgrade process. Not doing this could cause a gap in rule coverage ({pull}120906[#120906]).
+
+NOTE: If you're using the {es-sec} API to re-enable rules, refer to <<bulk-actions-rules-api, Bulk rule actions>>.
+
+To re-enable your rules:
+
+. Go to the All rules table on the Rules page (*Detect -> Rules*).
+. Click the checkbox in the *Rule* column to select all rules.
```
joepeeples, line 41: I think selecting this checkbox will only select the rules that are currently displayed on the page (20 rows by default). There's a separate UI element for "Select all ### rules".

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
 [discrete]
 [[known-issues-8.0.0-rc2]]
 ==== Known issues
```
joepeeples, line 10: Do we have (or want to have) some standardized wording that says that we plan to fix these issues in a future release?

joepeeples (commented):
A couple of comments, thanks!

docs/release-notes/8.0.asciidoc
nastasha-solomon: @jmikell821 and @benironside , @joepeeples noted that [in our docs](https://www.elastic.co/guide/en/security/current/rules-ui-management.html#select-all-prebuilt-rules-white_check_mark) that we use “click **Select all (number) rules**”. WDTY about continuing to use this when referring to the the **Select all rules** button on the Rules table?

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
...
+
+*The Data View option might not display in upgraded environments with legacy alerts:*
+
+NOTE: If new alerts are generated in an upgraded environment without legacy alerts, refreshing any page in {es-sec} will make the *Data View* option appear in the {es-sec-ui}.
+
+After upgrading, the *Data View* option might not display in environments with legacy alerts. To resolve this, a user with elevated role privileges must visit the {es-sec-app}, open a page that displays alert data (such as the Overview page), and refresh the page ({pull}121390[#121390]).
```
benironside, line 16: Since both lines 14 and 16 recommend that the user refresh to make the Data View button appear, would it be more concise to combine them and just say something like "After upgrading, the *Data View* option might not appear until a user with elevated privileges visits a page that displays alerts (such as the Overview page), and refreshes the page"?

I guess that would gloss over the detail currently in line 14 that in an env without legacy alerts, you can refresh any page rather than just one that shows alerts. But would it be common for users not to have any legacy alerts?

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
...
+** `.internal.alerts-security.alerts-<kib space>-*`
+
+* *{kib} space*: `All` privileges for the `Security` feature (see
+{kibana-ref}/xpack-spaces.html#spaces-control-user-access[Feature access based on user privileges])
+
+*Rules must be manually re-enabled after upgrading:*
+
+Rules are automatically disabled when upgrading to {stack} version 8.0.0 and must be manually re-enabled after completing the upgrade process. Not doing this could cause a gap in rule coverage ({pull}120906[#120906]).
+
+NOTE: If you're using the {es-sec} API to re-enable rules, refer to <<bulk-actions-rules-api, Bulk rule actions>>.
```
benironside, line 36:
```suggestion
NOTE: To use the {es-sec} API to re-enable rules, refer to <<bulk-actions-rules-api, Bulk rule actions>>.
```

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
...
+
+To re-enable your rules:
```
benironside, line 38:
```suggestion
To re-enable your rules from the Rules page:
```
My thinking here is to contrast this option with the API approach mentioned in the note above.

Or maybe, "To re-enable your rules from the app", or "To re-enable your rules manually".

benironside (commented):
LGTM, left a couple suggestions for your consideration.

docs/release-notes/8.0.asciidoc
nastasha-solomon: This is great suggestion!

docs/release-notes/8.0.asciidoc
nastasha-solomon: Thanks for this feedback! I don't know how common it'd be for users to not have legacy alerts in their environment, but that'd be the case if they started directly from `8.0` or decided to not migrate their old alerts from `7.x`. 

I'm also glad you commented on this because, as I was drafting the summary, I was wondering whether the placement of the note and the known issue description would be _too_ close. The wording in both is so similar that I could easily see a reader assuming they were the same. In reality, they describe two very different scenarios. 

**Scenario #1:**
Here, the user upgraded from `7.x` to `8.0` and, as part of the upgrade process, migrated alerts that were generated in `7.x`. Because of this, legacy alerts now exist in the upgraded environment (`8.0`) and in order to make the **Data View** option appear, a user with elevated role privs needs to visit the Security app and refresh a page that displays alert data.

**Scenario #2:**
The user did not upgrade from `7.x`, and just started using `8.0` instead. Because they didn’t upgrade from a previous version, they didn’t need to migrate any “old” alerts and the environment that they’re working in doesn’t contain legacy alerts. In this scenario, they still have to force the **Data View** option to appear, but they can only do it after new alerts are generated. Once new alerts are generated, the user can go to any page in the Security app and refresh it to force the **Data View** option to appear in the Security app.

One thing your comment made me realize was that it’s probably better to instruct the user to refresh a page that displays alert data, as opposed to any page in the Security app. That’s because, if they refresh a page that displays alert data (e.g., the Overview page), they can immediately verify that the **Data View** option is available for use. However, if they went to the Cases page and refreshed that, they wouldn’t be able to immediately see that the **Data View** option is available because the Cases page only displays case data.

docs/release-notes/8.0.asciidoc
nastasha-solomon: Not that I'm aware of. I'm open to adding that language in if requested though!

docs/release-notes/8.0.asciidoc
jmikell821, line 16:
```suggestion
After upgrading, the *Data view* option might not display in environments with legacy alerts. To resolve this, a user with elevated role privileges must visit the {es-sec-app}, open a page that displays alert data (such as the Overview page), then refresh the page ({pull}121390[#121390]).
```

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
 [discrete]
 [[known-issues-8.0.0-rc2]]
 ==== Known issues
+
+*The Data View option might not display in upgraded environments with legacy alerts:*
```
jmikell821, line 12:
```suggestion
*The Data view option might not display in upgraded environments with legacy alerts:*
```

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
...
+
+NOTE: If new alerts are generated in an upgraded environment without legacy alerts, refreshing any page in {es-sec} will make the *Data View* option appear in the {es-sec-ui}.
```
jmikell821, line 14:
```suggestion
NOTE: If new alerts are generated in an upgraded environment without legacy alerts, refreshing any page in {elastic-sec} will make the *Data view* option appear in the {es-sec-ui}.
```

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
...
+** `.items-<kib-space>`
+** `.alerts-security.alerts-<kib-space>`
+** `.internal.alerts-security.alerts-<kib-space>-*`
+
+* *{kib} space*: `All` privileges for the `Security` feature (see
+{kibana-ref}/xpack-spaces.html#spaces-control-user-access[Feature access based on user privileges])
+
+*Rules must be manually re-enabled after upgrading:*
+
+Rules are automatically disabled when upgrading to {stack} version 8.0.0 and must be manually re-enabled after completing the upgrade process. Not doing this could cause a gap in rule coverage ({pull}120906[#120906]).
```
jmikell821, line 34: Just a suggestion here. 

```suggestion
Rules are automatically disabled when upgrading to {stack} version 8.0.0 and must be manually re-enabled after completing the upgrade process. Failure to do so could cause a gap in rule coverage ({pull}120906[#120906]).
```

jmikell821 (commented):
Just a few comments. Tagging a few additional reviewers as well.

peluja1012:
@marshallmain would it be better to recommend that users manually disable rules before upgrade?
image

marshallmain:
@peluja1012 @nastasha-solomon Yes, I think it would be good if we can frame the automatic disabling of rules as less ideal than users manually disabling the rules before upgrade.

nastasha-solomon:
@marshallmain @peluja1012 for #120906, how about something like:
If you do not manually disable rules before upgrading to Elastic Stack version 8.0.0, they're automatically disabled during the upgrade process and must be manually re-enabled after the process completes. Failure to do so could cause a gap in rule coverage

marshallmain:
That sounds good!

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
 [discrete]
 [[known-issues-8.0.0-rc2]]
 ==== Known issues
+
+*The Data View option might not display in upgraded environments with legacy alerts:*
+
+The *Data View* option might not display in environments with legacy alerts. To resolve this, a user with elevated role privileges must visit the {es-sec-app}, open a page that displays alert data (such as the Overview page), then refresh the page ({pull}121390[#121390]).
```
jmikell821, line 14: Should we repeat this if it's mentioned in the RN section title?

docs/release-notes/8.0.asciidoc
jmikell821, line 12: These are titles, so we can probably remove the colon. 

```suggestion
*The Data View option might not display in upgraded environments with legacy alerts*
```

jmikell821 (commented):
@nastasha-solomon a few small suggestions. Otherwise looks good.

docs/release-notes/8.0.asciidoc
nastasha-solomon: Good point, I don't think repeating it brings any extra value. How about removing it and revising the summary to:
To make the **Data View** option appear, a user with elevated role privileges must visit the {es-sec-app}, open a page that displays alert data (such as the Overview page), then refresh the page.

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
...
+** `.items-<kib-space>`
+** `.alerts-security.alerts-<kib-space>`
+** `.internal.alerts-security.alerts-<kib-space>-*`
+
+* *{kib} space*: `All` privileges for the `Security` feature (see
+{kibana-ref}/xpack-spaces.html#spaces-control-user-access[Feature access based on user privileges])
+
+NOTE: If new alerts are generated in an upgraded environment without legacy alerts, refreshing any page with alert data in {elastic-sec} will make the *Data view* option appear in the {es-sec-ui}.
+
+*Rules must be manually re-enabled after upgrading:*
```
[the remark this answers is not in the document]
nastasha-solomon: Hm...I think the issue here is that, after upgrading to 8.0, rules are not _actually_ enabled unless users manually enable them. @peluja1012 @marshallmain is that correct? If yes, how about revising the summary description to:
Incorrectly activated rules may not generate alerts after upgrading to {stack} 8.0.0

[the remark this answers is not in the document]
marshallmain: That's correct, after upgrading to 8.0 the rules will be disabled. Once the rules are activated they will begin generating alerts.

docs/release-notes/8.0.asciidoc
```diff
@@ -8,3 +8,35 @@
...
+** `.items-<kib-space>`
+** `.alerts-security.alerts-<kib-space>`
+** `.internal.alerts-security.alerts-<kib-space>-*`
+
+* *{kib} space*: `All` privileges for the `Security` feature (see
+{kibana-ref}/xpack-spaces.html#spaces-control-user-access[Feature access based on user privileges])
+
+NOTE: If new alerts are generated in an upgraded environment without legacy alerts, refreshing any page with alert data in {elastic-sec} will make the *Data view* option appear in the {es-sec-ui}.
+
+*Incorrectly activated rules may not generate alerts after upgrading to {stack} 8.0.0*
```
jmikell821, line 32: I think "incorrectly activated rules" potentially can be confusing. Let's just simplify a bit. 
```suggestion
*Detection rules may not generate alerts after upgrading to {stack} 8.0.0*
```

jmikell821 (commented):
@nastasha-solomon one small comment, then good to go.

stephmilovic (approved):
Permissions relating to Data Views LGTM! Thanks @nastasha-solomon

Issue #1402, opened 2022-01-13
[DOCS] UI updates to Case feature 


nastasha-solomon:
## Description 
Addresses https://github.com/elastic/kibana/pull/121876 and https://github.com/elastic/kibana/pull/121902. 

The following changes have been made to the Cases feature in `8.0`:
-  In the empty state (i.e., when no cases exist) the Cases page now shows a button with the text **Create case** (it used to be **Add New Case**). The message above the button has been revised as well. See https://github.com/elastic/kibana/issues/121767 for before and after images of the button and text.
- On the Create case page, numbers have been added to each step in workflow. 
<img width="1444" alt="new-numbers" src="https://user-images.githubusercontent.com/79124755/149414320-100218f1-35ec-4a8d-afa7-167218df8c3c.png">

## Docs
- [Cases](https://www.elastic.co/guide/en/security/master/cases-overview.html)

PR #1442, opened 2022-01-25
[DOCS] UI updates to Case feature

nastasha-solomon:
Addresses https://github.com/elastic/security-docs/issues/1402. 

Preview [here](https://security-docs_1442.docs-preview.app.elstc.co/guide/en/security/master/cases-overview.html#cases-ui-open).


docs/cases/cases-overview.asciidoc
```diff
@@ -31,7 +31,11 @@ image::images/cases-home-page.png[Case UI Home]
 Open a new case to keep track of security issues and share their details with
 colleagues.
 
-. Go to *Cases* -> *Create new case*.
+. Go to the *Cases* page and click *Create case*. If no cases exist, the Cases table will be empty and you'll be prompted to create one by clicking the *Create case* button inside the table.
```
joepeeples, line 34: On other pages, the format would be more like this. But the way you've written it is also perfectly readable, so just pointing this out for possible consistency.
```suggestion
. Go to *Detect* -> *Cases*, then click *Create case*. If no cases exist, the Cases table will be empty and you'll be prompted to create one by clicking the *Create case* button inside the table. 
```

joepeeples (approved):
Just one comment, which you can take or leave. 😄 LGTM!

docs/cases/cases-overview.asciidoc
nastasha-solomon: I actually like this better _because_ it's consistent with the format our docs typically follow!

docs/cases/cases-overview.asciidoc
```diff
@@ -31,7 +31,11 @@ image::images/cases-home-page.png[Case UI Home]
 Open a new case to keep track of security issues and share their details with
 colleagues.
 
-. Go to *Cases* -> *Create new case*.
+. Go to *Detect* -> *Cases*, then click *Create case*. If no cases exist, the Cases table will be empty and you'll be prompted to create one by clicking the *Create case* button inside the table.
```
benironside, line 34:
```suggestion
. Go to *Detect* -> *Cases*, then click *Create case*. If no cases exist, you can create one by clicking the *Create case* button inside the table. 
```
I think we probably don't need to tell them the cases table will be empty since they're sure to see that. In fact maybe we don't even need the second sentence in this line since whether or not there are pre-existing cases, they can still click the same Create case button.

jmikell821: +1 for Ben's suggestion.

jmikell821 (approved):
Once you address the one comment Ben gave, this'll be good to go.

docs/cases/cases-overview.asciidoc
nastasha-solomon: Yeah, I debated on describing the table's empty state. I agree that it doesn't add much value since, like you said, users can click on the button above the table regardless of whether cases exist. The button name in the empty table is the same too, so although the workflow to access the button is different, users are essentially clicking on the same button. IMO, the screenshot isn't completely necessary but it wouldn't hurt to keep either. WDYT @benironside @jmikell821 ?

cnasikas: I think cases are under the `Investigate` section and not under the `Detect`

Screenshot 2022-02-01 at 12 32 25 PM

nastasha-solomon: That they are, good catch - thanks! I'll make this change right now.

cnasikas (approved):

PR #1456, opened 2022-01-27
[DOCS] Alert schema 8.0

### LINKED ISSUE #1385: [DOCS] Add New section for Security Solution Alert schema to `Elastic Security fields and object schemas` in 8.0
Related issue: https://github.com/elastic/kibana/pull/121135

## Description 

Users that consume data from the .siem-signals index will need to know that in 8.0 we've changed the schema used for Alert documents used in the Security Solution.
image

*Additional Details*

- The Alert Schema contains both ECS and Non-ECS fields
- The non-ECS field mapping is described at:  https://github.com/elastic/kibana/blob/main/x-pack/plugins/security_solution/server/lib/detection_engine/routes/index/signal_aad_mapping.json 

## Acceptance Test Criteria

- [ ] Users that access documents in the .siem-signals-* indices, for example to implement private security operations workflows, after viewing this page, will be able to modify their API queries and scripts to operate properly on the new 8.0+ Alert documents.
- [ ] This document section provides a handy cross reference, listing old 7.x field names as well as 8.x field names.
- [ ] Links to the relevant ECS fields and field sets are included.
- [ ] The field specification shall contain the Beta disclaimer -  The schema is considered Beta and subject to change, even though the detection engine that creates these documents remains GA.

## Notes

- I've proposed some draft changes in a private Google Doc, provided to @jmikell821   
- The format of the docs could be the same as https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html 
- The Schema shall be considered Beta.

jmikell821:
Resolves #1385. 

[Alert schema preview](https://security-docs_1456.docs-preview.app.elstc.co/guide/en/security/master/alert-schema.html).


docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,118 @@
+[[alert-schema]]
+
+== Alert schema
+
+{elastic-sec} stores alerts that have been generated by detection rules in hidden {es} indices.  In 7.x versions, the index pattern was `.siem-signals-<spacename>-*`.  In 8.x versions, the index pattern is `.alerts-security.alerts-<spacename>-*`, and the field names have changed. The following table includes the new names and provides a cross-reference to the previous field names.
+
```
MikePaquette, line 6: We need to add a Beta Disclaimer here, applying to the non-ECS fields.  Something like "the non-ECS fields listed here are [beta] and subject to change."

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+|{ecs-ref}/ecs-url.html[`url.*`]|`url.*`|ECS `url.*` fields copied from the source document when available.
+|{ecs-ref}/ecs-user.html[`user.*`]|`user.*`| ECS `user.*` fields copied from the source document when available.
+|{ecs-ref}/ecs-user_agent.html[`user_agent.*`]|`user_agent.*`| ECS `user_agent.*` fields copied from the source document when available.
+|{ecs-ref}/ecs-vulnerability.html[`vulnerability.*`]|`vulnerability.*`|ECS `vulnerability.*` fields copied from the source document when available.
+|`signal.ancestors.*`|`kibana.alert.ancestors.*`|
+|`signal.depth`| `kibana.alert.depth`|
+|`signal.original_event.*` | `kibana.alert.original_event.*`|
+|`signal.original_time`|`kibana.alert.original_time`| The value copied from the source event (`@timestamp`).
+|`signal.reason` |`kibana.alert.reason`|
+|`signal.rule.author` |`kibana.alert.rule.author`|
```
MikePaquette, line 71: For all rows that start with `signal.rule.x` can we add a description like:
value of x from the rule that created this alert - see https://www.elastic.co/guide/en/security/current/rules-ui-create.html#rule-ui-advanced-params or https://www.elastic.co/guide/en/security/current/rules-ui-create.html#rule-ui-basic-params, depending on which matches?

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,118 @@
+[[alert-schema]]
+
+== Alert schema
+
+{elastic-sec} stores alerts that have been generated by detection rules in hidden {es} indices.  In 7.x versions, the index pattern was `.siem-signals-<spacename>-*`.  In 8.x versions, the index pattern is `.alerts-security.alerts-<spacename>-*`, and the field names have changed. The following table includes the new names and provides a cross-reference to the previous field names.
+
+NOTE: Users are advised NOT to use the `_source field` in alert documents, but rather use the fields API to programmatically obtain the list of fields used in these documents. Learn more about {ref}/search-fields.html[retrieving selected fields from a search].
+
+|==============================================
+|7.x signal field |8.x alert field | Description
```
MikePaquette, line 10: Can we add in a column for field datatype?

MikePaquette (changes requested):
Thanks @jmikell821 for putting this together!  I have made one change, and added a number of questions and suggestions.  I will be happy to work on further edits to the field details after we resolve these.  Thanks again!

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+|`signal.rule.to` |`kibana.alert.rule.to`|
+|`signal.rule.type` |`kibana.alert.rule.type`|
+|`signal.rule.updated_at` |`kibana.alert.rule.updated_at`|
+|`signal.rule.updated_by` |`kibana.alert.rule.updated_by`|
+|`signal.rule.version` |`kibana.alert.rule.version`|
+|`signal.status` |`kibana.alert.workflow_status`|
+|`signal.threshold_result.*`|`kibana.alert.threshold_result.*`|
+|`signal.group.id` |`kibana.alert.group.id`|
+|`signal.group.index` |`kibana.alert.group.index`|
+|`signal.rule.index` |`kibana.alert.rule.parameters.index`|
```
marshallmain, line 104: @MikePaquette do you think we should document the fields within `kibana.alert.rule.parameters` publicly? These fields are copied directly from the rule as a record of the rule at the time of execution. Perhaps we could document `kibana.alert.rule.parameters` as a single field and say something like "Contains the rule parameters at the time the alert was generated" but not document individual fields within `parameters`.

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+|{ecs-ref}/ecs-allowed-values-event-type.html[`event.type`]|`event.type`| ECS field, copied from the source document when available.
+|{ecs-ref}/ecs-allowed-values-event-outcome.html[`event.outcome`]|`event.outcome` | ECS field, copied from the source document when available.
+|{ecs-ref}/ecs-agent.html[`agent.*`]| `agent.*`| ECS `agent.*` fields copied from the source document when available.
+|{ecs-ref}/ecs-client.html[`client.*`] | `client.*` | ECS `client.*` fields copied from the source document when available.
+|{ecs-ref}/ecs-cloud.html[`cloud.*`] |`cloud.*` | ECS `cloud.*` fields copied from the source document when available.
+|{ecs-ref}/ecs-container.html[`container.*`] | `container.*` | ECS `container.* fields` copied from the source document when available.
+|{ecs-ref}/ecs-data_stream.html[`data_stream.*`] |`data_stream.*`| ECS `data_stream.*` fields copied from the source document when available.
+
+NOTE: These fields may be constant keywords in the source documents, but they need to be copied into the alerts documents as keywords.
```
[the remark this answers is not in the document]
jmikell821: Hey @marshallmain, not sure? But it was referenced in your [Google sheet](https://docs.google.com/spreadsheets/d/1Cqv0bRx1BgY6VP9PY4FK6kppUEZJ5qMPoFgr5_XQXJI/edit?usp=sharing), line 23. Should I remove this sentence?

docs/reference/alert-schema.asciidoc
jmikell821: Hey @MikePaquette, what would go in that field datatype column? I need to be mindful of how many columns are in the table b/c due to some styling limitations beyond my control, the table will start to look skewed if it's too wide. It can be as long as needed, but 5 columns (currently 3 + the 2 you suggested) may cause some visual issues. The table already isn't wrapping the way I'd ideally like it to b/c of the long field names in the second column.

docs/reference/alert-schema.asciidoc
jmikell821: @MikePaquette I'm not quite sure I understand 100%. If I put a description of `signal.rule.building_block_type` according to this ^, wouldn't it read something like "The value of `building_block_type` from the rule that created this alert." Is that format correct for all the `signal.rule.x` fields?

docs/reference/alert-schema.asciidoc
MikePaquette: I understand the challenges with adding more columns.  The field datatype is the Elasticsearch datatype used by the index mapping template for the alerts indices.  In other words, what kind of field is this (keyword, multi-field, long, date, boolean, etc.). If we can't add a column, perhaps we can add this in the description.  It would be necessary only for the non-ECS fields.  Note:  the ECS field reference includes this information in the text of the description like this:  
image

docs/reference/alert-schema.asciidoc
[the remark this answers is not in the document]
MikePaquette: Ah yes, the internal spreadsheet was targeted at the developers of alerts (Elastic folks), while this document is for consumers of alerts (users), so we need to change the note.  There is no action required of users.

So if we add the field datatype to these definitions per the comment at row 8 above, this field would have "type: keyword", and we could add the note that says "Note: These fields may be constant keywords in the source documents, but they are copied into the alerts documents as keywords."

docs/reference/alert-schema.asciidoc
MikePaquette: I like the idea, but some of these fields are not documented anywhere else.  For example building_block_type, to, from, timeline_id, timeline_title, max_signals, so I think it's best if this document serves as the place where these are defined for now.  So my preference is that we keep all of them here and add relevant definitions where they are not covered by the rule basic and advanced settings docs.

docs/reference/alert-schema.asciidoc
MikePaquette: Good question. there are indeed a few fields, including `building_block_type` that are different from most of the others that start with `signal.rule.x`.  I see building_block_type, to, from, timeline_id, timeline_title, max_signals, that should be treated differently.  Let's treat them as special cases, and add a full definition of the field here, rather than pointing to the basic and advance settings.

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,120 @@
+[[alert-schema]]
+
+== Alert schema
+
+{elastic-sec} stores alerts that have been generated by detection rules in hidden {es} indices.  In 7.x versions, the index pattern was `.siem-signals-<spacename>`.  In 8.x versions, the index pattern is `.alerts-security.alerts-<spacename>`, and the field names have changed. The following table includes the new names and provides a cross-reference to the previous field names.
+
+NOTE: Users are advised NOT to use the `_source field` in alert documents, but rather use the fields API to programmatically obtain the list of fields used in these documents. Learn more about {ref}/search-fields.html[retrieving selected fields from a search].
```
benironside, line 7:
```suggestion
NOTE: Users are advised NOT to use the `_source field` in alert documents, but rather to use the fields API to programmatically obtain the list of fields used in these documents. Learn more about {ref}/search-fields.html[retrieving selected fields from a search].
```

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,120 @@
...
+|{ecs-ref}/ecs-allowed-values-event-outcome.html[`event.outcome`]|`event.outcome` | ECS field, copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-agent.html[`agent.*`]| `agent.*`| ECS `agent.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-client.html[`client.*`] | `client.*` | ECS `client.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-cloud.html[`cloud.*`] |`cloud.*` | ECS `cloud.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-container.html[`container.*`] | `container.*` | ECS `container.* fields` copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-data_stream.html[`data_stream.*`] |`data_stream.*`| ECS `data_stream.*` fields copied from the source document, if present, for custom query and indicator match rules.
+
+NOTE: These fields may be constant keywords in the source documents, but they need to be copied into the alerts documents as keywords.
```
benironside, line 28:
```suggestion
NOTE: These fields may be constant keywords in the source documents, but are copied into the alerts documents as keywords.
```

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,120 @@
...
+|{ecs-ref}/ecs-organization.html[`organization.*`]| `organization.*`| ECS `organization.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-package.html[`package.*`]|`package.*`|ECS `package.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-process.html[`process.*`]| `process.*`| ECS `process.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-registry.html[`registry.*`] | `registry.*`| ECS `registry.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-related.html[`related.*`] |`related.*` |ECS `related.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-rule.html[`rule.*`]| `rule.*`| ECS `rule.*` fields copied from the source document, if present, for custom query and indicator match rules.
+
+NOTE: These fields are not related to the detection rule that generated the alert.
```
benironside, line 53: It's not obvious to me whether this references the previous or the following fields, especially since the "Description" column values are so similar before and after the note. Could instead say "The preceding" or "The following" fields.

benironside (commented):
Left a few minor suggestions. Bigger picture, it seems to me that the description fields could be significantly improved, which will likely be a substantial project. I mean that while the field names themselves are pretty descriptive, most of the descriptions don't explicitly say what the field represents, or what values it might contain (or in what format those values appear, e.g. numeric, string, boolean). Most of the existing descriptions are identical and could seemingly be replaced by a note saying something like "The following fields, if present in the source document, are copied to the alert." 

Maybe we should create an issue to revisit this page and update the descriptions?

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,120 @@
+[[alert-schema]]
+
+== Alert schema
+
+{elastic-sec} stores alerts that have been generated by detection rules in hidden {es} indices.  In 7.x versions, the index pattern was `.siem-signals-<spacename>`.  In 8.x versions, the index pattern is `.alerts-security.alerts-<spacename>`, and the field names have changed. The following table includes the new names and provides a cross-reference to the previous field names.
```
joepeeples, line 5: It seems a little weird to lead with the 7.x info everywhere on this page, since these are the 8.0 docs. I know that a lot of customers will be upgrading from 7.x, but it could be distracting for new customers starting on 8.x. Also, why don't the 8.x field names have links like a lot the 7.x fields? Again, that seems to privilege 7.x.

This suggestion reorders to focus on 8.x first, and also specifies that _some_ fields are different, not all.

```suggestion
{elastic-sec} stores alerts that have been generated by detection rules in hidden {es} indices. In 8.x versions, the index pattern is `.alerts-security.alerts-<spacename>`. In 7.x versions, the index pattern was `.siem-signals-<spacename>`, and some field names were different. The following table includes the current names and cross-references the legacy field names.
```

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,120 @@
...
+|{ecs-ref}/ecs-tracing.html[`trace.*`]| `trace.*`|ECS `trace.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-tracing.html#field-transaction-id[`transaction.*`]|`transaction.*`| ECS `transaction.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-url.html[`url.*`]|`url.*`|ECS `url.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-user.html[`user.*`]|`user.*`| ECS `user.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-user_agent.html[`user_agent.*`]|`user_agent.*`| ECS `user_agent.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-vulnerability.html[`vulnerability.*`]|`vulnerability.*`|ECS `vulnerability.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|`signal.ancestors.*`|`kibana.alert.ancestors.*`|
```
joepeeples, line 67: Move all of these signal rules up in the table so they're in alphabetical order with the rest of the list.

docs/reference/alert-schema.asciidoc
joepeeples, line 7: This implies that there is something called "the fields API," which I don't think is quite right. This is more consistent with the info on the linked page:
```suggestion
NOTE: Users are advised NOT to use the `_source` field in alert documents, but rather to use the `fields` option on the search API to programmatically obtain the list of fields used in these documents. Learn more about {ref}/search-fields.html[retrieving selected fields from a search].
```

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,120 @@
+[[alert-schema]]
+
+== Alert schema
+
+{elastic-sec} stores alerts that have been generated by detection rules in hidden {es} indices.  In 7.x versions, the index pattern was `.siem-signals-<spacename>`.  In 8.x versions, the index pattern is `.alerts-security.alerts-<spacename>`, and the field names have changed. The following table includes the new names and provides a cross-reference to the previous field names.
+
+NOTE: Users are advised NOT to use the `_source field` in alert documents, but rather use the fields API to programmatically obtain the list of fields used in these documents. Learn more about {ref}/search-fields.html[retrieving selected fields from a search].
+The non-ECS fields listed below are [beta] and subject to change.
```
joepeeples, line 8: Should this line be a separate paragraph? It renders as part of the previous paragraph, which kind of loses impact. Convert to a [note block](https://github.com/elastic/docs#admon-blocks)?

Also, should it be `beta::[]` to render with the beta admonition?

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,120 @@
...
+|{ecs-ref}/ecs-user_agent.html[`user_agent.*`]|`user_agent.*`| ECS `user_agent.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-vulnerability.html[`vulnerability.*`]|`vulnerability.*`|ECS `vulnerability.*` fields copied from the source document, if present, for custom query and indicator match rules.
+|`signal.ancestors.*`|`kibana.alert.ancestors.*`|
+|`signal.depth`| `kibana.alert.depth`|
+|`signal.original_event.*` | `kibana.alert.original_event.*`|
+|`signal.original_time`|`kibana.alert.original_time`| The value copied from the source event (`@timestamp`).
+|`signal.reason` |`kibana.alert.reason`|
+|`signal.rule.author` |`kibana.alert.rule.author`| The value of the `author` who created the rule. See <<rule-ui-advanced-params, configure advanced rule settings>>.
+|`signal.rule.building_block_type`|`kibana.alert.building_block_type`|The value of `building_block_type` from the rule that generated this alert. See <<rule-ui-advanced-params, configure advanced rule settings>>.
```
joepeeples, line 73: "Refer to" probably works here too:
```suggestion
|`signal.rule.author` |`kibana.alert.rule.author`| The value of the `author` who created the rule. Visit <<rule-ui-advanced-params, configure advanced rule settings>>.
|`signal.rule.building_block_type`|`kibana.alert.building_block_type`|The value of `building_block_type` from the rule that generated this alert. Visit <<rule-ui-advanced-params, configure advanced rule settings>>.
```

joepeeples (commented):
Added a few comments and piggybacked on others' suggestions. I really like @MikePaquette's idea of a separate column for "ECS Field" to avoid repeating that in the Description, but the current docs system probably prevents adding more columns.

docs/reference/alert-schema.asciidoc
```diff
@@ -0,0 +1,120 @@
...
+The non-ECS fields listed below are [beta] and subject to change.
+
+|==============================================
+|7.x signal field |8.x alert field | Description
+
+|{ecs-ref}/ecs-base.html#field-timestamp[`@timestamp`] |`@timestamp` | ECS field, represents the time when the alert was created or most recently updated.
+|{ecs-ref}/ecs-base.html#field-message[`message`]| `message` | ECS field copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-base.html#field-tags[`tags`]|`tags`| ECS field copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-base.html#field-labels[`labels`]| `labels`| ECS field copied from the source document, if present, for custom query and indicator match rules.
+|{ecs-ref}/ecs-ecs.html#field-ecs-version[`ecs.version`] |`ecs.version`| ECS field, should NOT be copied from source documents.
```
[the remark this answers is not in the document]
jmikell821: @marshallmain or @MikePaquette checking to see what recommendation you have here ^ or if this line is correct.

docs/reference/alert-schema.asciidoc
jmikell821: @benironside this is in the same table cell, so it's pertaining to line 51.

docs/reference/alert-schema.asciidoc
jmikell821: Added `Type` in the Description column.

marshallmain:
The page looks accurate to me for converting from 7.x field names to 8.0 field names. There are a few fields that are new to the alert schema in 8.0 and aren't represented here:
- `kibana.space_ids`
- `kibana.alert.rule.consumer`
- `kibana.alert.status` (different from `kibana.alert.workflow_status`, which is more important to security alerts and is represented here already :+1: )
- `kibana.alert.rule.category`
- `kibana.alert.rule.execution.uuid`
- `kibana.alert.rule.producer`
- `kibana.alert.rule.rule_type_id`

Is this page intended to focus on how existing fields in 7.x are changing in 8.0, or is it attempting to capture the full 8.0 schema in addition to the way fields are changing from 7.x to 8.0? I think it makes sense to have a page that focuses only on the conversion from 7 to 8, but I wanted to confirm if that's the intent here.

docs/reference/alert-schema.asciidoc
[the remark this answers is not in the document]
MikePaquette: I agree.  Let's change as recommended by @benironside and @marshallmain

MikePaquette:
@marshallmain @jmikell821 

> Is this page intended to focus on how existing fields in 7.x are changing in 8.0, or is it attempting to capture the full 8.0 schema in addition to the way fields are changing from 7.x to 8.0? I think it makes sense to have a page that focuses only on the conversion from 7 to 8, but I wanted to confirm if that's the intent here.

Currently this page is attempting to capture the full 8.0 schema in addition to the way fields are changing from 7.x to 8.0, so 
we should add these fields to the document.

Regarding how to add them, I am fine with new section, but I think it might be best to just add them to the existing table with a "none" or "not used" or "N/A" in the signals column.    @marshallmain can you provide the datatype for each of these 7 fields so it can be included in the description column?

marshallmain:
I updated my comment above with types for each new field.

Issue #1635, opened 2022-03-01
[DOCS] Clarify note about exported case alerts 


nastasha-solomon:
## Description 

In [our docs](https://www.elastic.co/guide/en/security/master/cases-overview.html#cases-export) for exporting a case, we tell users that exported alerts contain metadata that allow the case to reference the appropriate alerts after they’re imported. 

<img width="690" alt="Screen Shot 2022-03-01 at 12 32 17 PM" src="https://user-images.githubusercontent.com/79124755/156219051-91692c18-bb44-4d98-a5f6-551c3753bb46.png">

There was a question in a detections Slack channel where the poster mentioned that, after importing a case with alerts, they weren't able to access alert details and weren't sure why. @spong clarified that, when importing a case, the raw alert doc _is not_ imported to the cluster and this is why data in the Alerts Details flyout and Rule Details link won't work as they normally do (i.e. the data won't fully display). 

In its current state, the note doesn't fully communicate these dependencies, so it might be worth updating to clarify additional actions users need to take after importing a case. 

cc: @cnasikas

PR #1622, opened 2022-02-28
[DOCS] KPIs added to Cases #1404

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/1404.

Preview:
- [Cases | Review the case summary](https://security-docs_1622.docs-preview.app.elstc.co/guide/en/security/master/cases-overview.html#cases-summary)
**NOTE:** I also made various updates to the Cases topic to refresh screenshots and remove instances of "view".
- [Elastic UI | Cases page](https://security-docs_1622.docs-preview.app.elstc.co/guide/en/security/master/es-ui-overview.html#cases-ui)
- [Manage detection alerts | Add an alert to an existing case](https://security-docs_1622.docs-preview.app.elstc.co/guide/en/security/master/alerts-ui-manage.html#signals-to-existing-cases)

PR #1637, opened 2022-03-02
[DOCS] Clarify note about exported case alerts 

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/1635.

Preview [here](https://security-docs_1637.docs-preview.app.elstc.co/guide/en/security/master/cases-overview.html#cases-import) (see the admonition under step 5).


docs/cases/cases-overview.asciidoc
```diff
@@ -118,9 +118,9 @@ IMPORTANT: Before importing Lens visualizations, Timelines, or alerts into a spa
 [float]
 [[cases-export]]
 === Export a case
-Use the export option to move cases between different Kibana instances. When you export a case, the following data is exported to a newline-delimited JSON (NDJSON) file: case details, user actions, text string comments, case alerts, and lens visualizations.
+Use the export option to move cases between different Kibana instances. When you export a case, the following data is exported to a newline-delimited JSON (`.ndjson`) file: case details, user actions, text string comments, case alerts, and lens visualizations (which are exported as JSON blobs).
 
-NOTE: Exported alerts contain metadata that allow the case to reference the appropriate alerts after they're imported. Lens visualizations are exported as JSON blobs.
+IMPORTANT: Alerts attached to a case are exported as references containing alert metadata. When a <<cases-import, case is imported>>, it uses these references to re-connect to the appropriate alerts. However, alert references only resolve if the correct alert documents are available in the environment. If they're not, alert details will not load when you access them from the imported case.
```
cnasikas, line 123: Thank you @nastasha-solomon! This is much better. Maybe we can add at the end "The same applies to the rules from which the alerts have been created" or similar 🙂

cnasikas (approved):

docs/cases/cases-overview.asciidoc
```diff
@@ -118,9 +118,9 @@ IMPORTANT: Before importing Lens visualizations, Timelines, or alerts into a spa
...
+Use the export option to move cases between different Kibana instances. When you export a case, the following data is exported to a newline-delimited JSON (`.ndjson`) file: case details, user actions, text string comments, case alerts, and lens visualizations (which are exported as JSON blobs).
 
-NOTE: Exported alerts contain metadata that allow the case to reference the appropriate alerts after they're imported. Lens visualizations are exported as JSON blobs.
+IMPORTANT: Alerts attached to a case are exported as references containing alert metadata. When a <<cases-import, case is imported>>, it uses these references to re-connect to the appropriate alerts. However, alert references only resolve if the correct alert documents are available in the environment. If they're not, alert details will not load when you access them from the imported case. Links to the rules that produced the alert are treated the same way; they only resolve if the rule exists in the environment.
```
spong, line 123: > Links to the rules that produced the alert are treated the same way; they only resolve if the rule exists in the environment.

So there's a slight nuance here with regards to links to Rule Details as we do have support for showing 'Deleted Rules', which this would classify as (even though it never existed). For this functionality (most recently updated in https://github.com/elastic/kibana/pull/120053), if a Rule doesn't exist, we'll try to show its details from the source Alert (by means of the RuleId which should stay static between Rule imports/exports). So if they re-import the associated Alert, links to Rule Details should still function, it'll just have a badge next to the name saying it was deleted.

spong (approved):
One small clarification around Rule Details links, but other than that LGTM! Thanks @nastasha-solomon! 🙂

docs/cases/cases-overview.asciidoc
```diff
@@ -118,9 +118,9 @@ IMPORTANT: Before importing Lens visualizations, Timelines, or alerts into a spa
...
+Use the export option to move cases between different Kibana instances. When you export a case, the following data is exported to a newline-delimited JSON (`.ndjson`) file: case details, user actions, text string comments, case alerts, and lens visualizations (which are exported as JSON blobs).
 
-NOTE: Exported alerts contain metadata that allow the case to reference the appropriate alerts after they're imported. Lens visualizations are exported as JSON blobs.
+IMPORTANT: Alerts attached to a case are exported as references containing alert metadata. When a <<cases-import, case is imported>>, it uses these references to re-connect to the appropriate alerts. However, alert references only resolve if the correct alert documents are available in the environment. If they're not, alert details will not load when you access them from the imported case. Links to the rule that produced the alert will resolve if the alert is present in the environment.
```
joepeeples, line 123: The first sentence uses "attach" and that seems like a good way to describe the relationship between alerts and cases. Also, "correct" was a little unclear (what makes an alert correct, or incorrect?).
```suggestion
IMPORTANT: Alerts attached to a case are exported as references containing alert metadata. When a <<cases-import, case is imported>>, it uses these references to reattach the appropriate alerts. However, alert references only resolve if the referenced alert documents are available in the environment. If they're not, alert details will not load when you access them from the imported case. Links to the rule that produced the alert will resolve if the alert is present in the environment.
```

joepeeples (commented):
A couple of word choice suggestions, thanks!

docs/cases/cases-overview.asciidoc
nastasha-solomon: Agree that "referenced alerts" is less ambiguous than "correct alerts" in this context. I'm not sure whether "reattach" is more suitable than "reconnect" though. @spong from your perspective, are they basically communicating the same idea, or is one more accurate than the other?

spong: Perhaps `find`, `locate`, or `look up` would be a suitable replacement for attach/connect here, as there is no technical re-association between the case/alert on import, but rather there's just a broken reference when trying to use features requiring the alert document(s) and it isn't available. You would see the same behavior if the alert document was deleted (and the case was never exported/imported). Does that make sense?

So maybe something along the lines:

```suggestion
IMPORTANT: Alerts attached to a case are exported as references containing alert metadata, so if a <<cases-import, case is imported>> into a cluster without the attached alerts, features relying on the full alert document will not function (e.g. viewing Alert Details, Rule Details, or syncing Alert status with the Case).
```

<p align="center">
  
</p>

cnasikas, line 123: > Links to the rule that produced the alert will resolve if the alert is present in the environment

This is not correct. Links to the rule will resolve only if the **rule** is present in the environment, not the alert. Cases saved two references, one for rules and one for alerts. They are autonomous and they did not depend on each other.

spong: @cnasikas -- while Cases only saves the rule reference, the `Rule Details` page supports showing 'deleted rules' (in this case 'missing') so long as an alert exists with the Rule's corresponding `id` (7.x) or `rule_id` (8.x). Please see this comment https://github.com/elastic/security-docs/pull/1637#discussion_r817824441 from @nastasha-solomon for details.

Even if the Rule were to be exported and re-imported alongside the Case, it'll have generated a new `id` on import (not `rule_id` which remains static), and so the Rule Details link from within Cases wouldn't find the Rule since it would be referencing the old `id`. I just tested this by creating a Rule, generating Alerts, creating a Case from the Alert, exporting the Rule, deleting the Rule, importing the Rule, and verifying the Rule link within the Case goes to the old ID which shows the `Deleted Rule` as it was recreated from the most recent Alert that was created before it was deleted.

Case
<p align="center">
  
</p>

Deleted Rule
<p align="center">
  
</p>

Re-imported Rule

<p align="center">
  
</p>

docs/cases/cases-overview.asciidoc
cnasikas: Thank you so much @nastasha-solomon and @spong! I wasn't aware of such functionality. Today I learned something 🙂.

docs/cases/cases-overview.asciidoc
cnasikas: Thank you @spong! I wasn't aware that you can view deleted rules if the alert exists.

benironside (approved):

docs/cases/cases-overview.asciidoc
```diff
@@ -118,9 +118,7 @@ IMPORTANT: Before importing Lens visualizations, Timelines, or alerts into a spa
 [float]
 [[cases-export]]
 === Export a case
-Use the export option to move cases between different Kibana instances. When you export a case, the following data is exported to a newline-delimited JSON (NDJSON) file: case details, user actions, text string comments, case alerts, and lens visualizations.
-
-NOTE: Exported alerts contain metadata that allow the case to reference the appropriate alerts after they're imported. Lens visualizations are exported as JSON blobs.
+Use the export option to move cases between different Kibana instances. When you export a case, the following data is exported to a newline-delimited JSON (`.ndjson`) file: case details, user actions, text string comments, case alerts, and lens visualizations (which are exported as JSON blobs).
```
jmikell821, line 121: Should "export" be in quotes or bold if it's an option in the UI?

docs/cases/cases-overview.asciidoc
```diff
@@ -147,4 +145,13 @@ To import a case:
 . Click *Import*.
 . Review the import log and click *Done*.
 +
-IMPORTANT: If you imported a case with a connector, you'll be prompted to re-authenticate the connector. To do so, click the *Go to connectors* button on the *Import saved objects* flyout and complete the necessary steps to re-authenticate the connector. Another way to access connectors is to open the main menu, then go to *Stack Management -> Alerts and Insights -> Rules and Connectors -> Connectors*.
+[IMPORTANT]
+=========================
+
+Be mindful of the following items:
```
jmikell821, line 151:
```suggestion
Be mindful of the following:
```

docs/cases/cases-overview.asciidoc
```diff
@@ -147,4 +145,13 @@ To import a case:
...
+
+* If the imported case had connectors attached to it, you'll be prompted to re-authenticate the connectors. To do so, click the *Go to connectors* button on the *Import saved objects* flyout and complete the necessary steps. Alternatively, open the main menu, then go to *Stack Management -> Alerts and Insights -> Rules and Connectors -> Connectors* to access connectors.
```
jmikell821, line 153:
```suggestion
* If the imported case had connectors attached to it, you'll be prompted to re-authenticate the connectors. To do so, click *Go to connectors* on the *Import saved objects* flyout and complete the necessary steps. Alternatively, open the main menu, then go to *Stack Management -> Alerts and Insights -> Rules and Connectors -> Connectors* to access connectors.
```

jmikell821 (commented):
Very small nits. You can merge afterwards - thanks!

docs/cases/cases-overview.asciidoc
nastasha-solomon: Yep, the button is actually named Export so I'll bold it here.

nastasha-solomon:
Removed the image to align the topic with https://github.com/elastic/security-docs/pull/1622.

Issue #1694, opened 2022-03-09
[BUG] Example response in "Find Cases by Alert" topic is outdated


nastasha-solomon:
The example response for the API that finds cases by alerts is outdated. With https://github.com/elastic/kibana/pull/102664, the case title is now returned along with--and separate from--the alert ID. #102664 was introduced in `7.14`, so will need to backport this change from `8.1`-`7.14`.

**Current:**

```["8af6ac20-74f6-11ea-b83a-553aecdb28b6", "a18b38a0-71b0-11ea-a0b2-c51ea50a58e2"]```

**Correct:**

```
[
    {
        "id": "8af6ac20-74f6-11ea-b83a-553aecdb28b6",
        "title": "Test 1"
    },
    {
        "id": "a18b38a0-71b0-11ea-a0b2-c51ea50a58e2",
        "title": "Test 2"
    }
]
```

cc: @cnasikas

PR #1695, opened 2022-03-09
[BUG] Example response in "Find Cases by Alert" topic is outdated

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/1694.

Preview [here](https://security-docs_1695.docs-preview.app.elstc.co/guide/en/security/master/cases-api-find-cases-by-alert.html). (Fixed outdated example and made some minor edits throughout the topic.)


docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
```diff
@@ -1,33 +1,33 @@
 [[cases-api-find-cases-by-alert]]
-=== Find Cases by Alert
+=== Find cases by alert
 
-Retrieves all case IDs that have a specific alert attached to the cases.
+Uses an alert ID (`alert_id`) to retrieve case IDs of the cases the alert is attached to.
```
benironside, line 4:
```suggestion
Uses an alert ID (`alert_id`) to retrieve the case IDs of any cases the alert is attached to.
```

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
```diff
@@ -1,33 +1,33 @@
...
 
-NOTE: Console supports only Elasticsearch APIs. Console doesn't allow interactions with {kib} APIs. You must use `curl` or another HTTP tool instead. For more information, refer to {kibana-ref}/console-kibana.html[Run {es} API requests].
+TIP: The case ID is identical to the case's saved object ID.
```
benironside, line 6:
```suggestion
TIP: The case's case ID is identical to its saved object ID.
```

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
```diff
@@ -42,11 +42,20 @@ GET api/cases/alerts/293f1bc0-74f6-11ea-b83a-553aecdb28b6
 
 ==== Response payload
 
-A JSON array listing the retrieved case IDs.
+A JSON array containing case IDs associated with the specified alert. Case titles are also retrieved.
```
benironside, line 45:
```suggestion
A JSON array containing any case IDs associated with the specified alert, and case titles.
```

benironside (commented):
LGTM, left a few suggestions for your consideration

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
nastasha-solomon: @benironside what are your thoughts on shortening it to just "case ID" instead of using "case's case ID"?

joepeeples: My $0.02: I lean toward "The case ID" -- shorter, less repetitive.

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
```diff
@@ -1,33 +1,33 @@
...
 
 ==== Request URL
 
-`GET <kibana host>:<port>/api/cases/alerts/<alert ID>`
+`GET <kibana host>:<port>/api/cases/alerts/<alert_id>`
 
 ===== URL parts
 
-The URL must include the `alert ID` of the alert that you are interested in find the associated cases it is attached to.
+The URL must include the ID of the alert that is attached to the case.
```
joepeeples, line 16: We might want to keep the specific field name (`alert ID`) here. It seems quicker to understand at a glance.

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
jmikell821: +1 for shortening to `case ID`.

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
joepeeples: Amendment to avoid a comma splice:
```suggestion
A JSON array containing any case IDs and case titles associated with the specified alert.
```

joepeeples (commented):
A few comments but LGTM!

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
jmikell821: `"and case titles"` sounds a tad odd. Let's leave the original for now but +1 for the first suggestion.

jmikell821 (commented):
@nastasha-solomon can we get a member of the threat hunting team to review this too? (Never mind, I see you have Christos tagged).

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
nastasha-solomon: @joepeeples as in - The URL must include the `alert ID` attached to the case.

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
nastasha-solomon: Hm, I'm not entirely sure whether the alert ID is associated with the case title. @cnasikas what are your thoughts on the recommended revision? Is it an accurate statement?

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
joepeeples: Almost -- but the alert is attached to the case, not the alert ID. Maybe more like the original -- it's a little wordier but accurate:
```suggestion
The URL must include the `alert ID` of the alert attached to the case
```

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
```diff
@@ -1,33 +1,33 @@
...
 
 ==== Request URL
 
-`GET <kibana host>:<port>/api/cases/alerts/<alert ID>`
+`GET <kibana host>:<port>/api/cases/alerts/<alert id>`
 
 ===== URL parts
 
-The URL must include the `alert ID` of the alert that you are interested in find the associated cases it is attached to.
+The URL must include the `alert ID` of the alert that is attached to the case.
```
cnasikas, line 16: I think it should be something along the lines "of the alert that is attached to some/a case". The word "the case" makes me think that the alert should be attached to a specific case and not any case.

cnasikas (approved):

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
cnasikas: I think if we mention the cases IDs then we should mention the cases titles. The alert ID is not associated with the title. The case ID is associated with the title. The response contains all cases that the alert is attached. From the cases properties we choose to return the id and the title.

nastasha-solomon: Thanks, @cnasikas ! I'll leave the edit as it is then.

docs/cases/api/cases-api/cases-api-find-cases-by-alert.asciidoc
nastasha-solomon: Ah, good point. We definitely don't want to imply that the alert can only be attached to a single case. One alternative is to remove any mention of cases, for example, "The URL must include the `alert ID` of the attached alert." Another option is to rewrite the sentence to something like "The URL must include the alert ID (alert id), which is used to any cases the alert is attached to."

cnasikas: Whatever works best for you 🙂 ! Both of them, lgtm.

PR #1741, opened 2022-03-18
[DOCS] Host risk score updates 

### LINKED ISSUE #1436: [DOCS] Update to Host Risk Score docs
## Description 

We've made some updates to Host Risk Score in 8.1. These need to be reflected in our official documentation about the feature.

1. Host Risk Score is no more limited to detection rule alerts (it never was, but we realized this in 8.1), nor is it limited to open alerts. This needs to be updated [here](https://www.elastic.co/guide/en/security/current/whats-new.html#experimental-feature-7.16) and [here](https://www.elastic.co/guide/en/security/current/host-risk-score.html).

2. Host Risk Score is NOT limited to alerts with "open" status anymore.

3. We've added risk multipliers, which will boost the final risk score based on certain attributes of the host itself. For example, alert activity on a server calls for greater scrutiny than that on a laptop, so we  boost the host risk score by 1.5 times if the host is a server. This is done AFTER calculating the weighted sum of the time-corrected risks.

4. Users can now view host risk score data in a few more places other than the Host Overview and Alerts Flyout. @machadoum can provide details and screenshots.

5. We also want to callout the fact that views in the Security App might break if users don't upgrade the Host Risk Score artifacts in accordance with the Stack update. We've already [called this out in the GitHub docs](https://github.com/elastic/detection-rules/pull/1716), but a callout in the official documentation might also be useful.

Unrelated to 8.1 changes, but I just noticed that the second paragraph [here](https://www.elastic.co/guide/en/security/current/host-risk-score.html) has the following sentence:  

> A weighted sum of the individual time-corrected rule risks is then applied to the risk score.

This is a bit unclear. A weighted sum is not applied to the risk score, but is used to calculate the risk score.

jmikell821:
Resolves #1436. 

Preview [here](https://security-docs_1741.docs-preview.app.elstc.co/guide/en/security/master/host-risk-score.html).


ajosh0504 (approved):
Looks great! 🚢

docs/experimental-features/experimental-features-intro.asciidoc
```diff
@@ -1,5 +1,5 @@
 [[sec-experimental-intro]]
-= Experimental features
+= Technical review
```
joepeeples, line 2: I've seen this as "technical _preview_", per [these guidelines](https://github.com/elastic/docs#using-the-experimental-admonition), https://github.com/elastic/docs/issues/2305, etc.
```suggestion
= Technical preview
```

joepeeples (commented):
Changes look good! Just one comment about "technical preview" vs. "review".

docs/experimental-features/host-risk-score.asciidoc
```diff
@@ -23,12 +25,15 @@ The following table shows how risk levels are applied to a host, based on the no
...
 
-To deploy the Host Risk Score framework in your environment, follow https://github.com/elastic/detection-rules/blob/main/docs/experimental-machine-learning/host-risk-score.md[these steps]. These instructions also include steps to enable the `riskyHostsEnabled` feature flag.
+NOTE: Host risk score artifacts should be updated when you upgrade the {stack}. To do this, download a release bundle that's compatible with the new {stack} version and repeat all the steps referenced above. Failure to do so might cause views in the {security-app} to break.
 
-NOTE: To view host risk score data on the {elastic-sec} Overview page and the Alert details flyout, you must enable the feature flag. However, enabling the feature flag is NOT required to view the Lens dashboards.
+NOTE: To view host risk score data in the {security-app}, you must enable the feature flag. However, enabling the feature flag is NOT required to view the Lens dashboards.
```
nastasha-solomon, line 34: What do you think about linking to [this set of instructions](https://github.com/elastic/detection-rules/blob/main/docs/experimental-machine-learning/host-risk-score.md#8-enable-kibana-features) in this note. That way, we guide users towards the steps to enable the feature flag. 

Also, would it be better to define the feature flag here (`riskyHostsEnabled`) instead of first defining it in the following section?

nastasha-solomon (approved):
Left some minor suggestions, but overall LGTM!

benironside (commented):
Looks great, I left some suggestions for a paragraph about some data science stuff.

Issue #1814, opened 2022-04-05
[DOCS] Alert details show linked cases 


nastasha-solomon:
## Description 

Alert details now show whether the alert has been added to a case or cases. This enables analysts to understand if the alert is being actively triaged or investigated in a case, or multiple cases. 

If the alert has not been added to any cases, the message "This alert was found in 0 cases" displays. 
<img width="297" alt="alert-case-0" src="https://user-images.githubusercontent.com/79124755/161867830-6a4c782b-e950-49d1-b1e9-bc09c0debdc8.png">

Example of what the message will look like if the alert was attached to a single case. Note that the case name is hyperlinked in the message. When users click on the link, the case opens from the Cases page. 
<img width="400" alt="alert-case-1" src="https://user-images.githubusercontent.com/79124755/161867840-406520ad-36eb-4429-b6de-7981f6a62c3d.png">

If the alert has been added to multiple cases, the message changes to "This alert was found in **N** cases". The number in the message increments by 1 and the cases are written in a comma separated list. 
<img width="558" alt="alert-case-2" src="https://user-images.githubusercontent.com/79124755/161868087-b7ef7507-7306-45ed-9692-f92ccecc2cc0.png">

## Related dev issues and PRs
- https://github.com/elastic/kibana/pull/128033

## Notes
- Note limit to number of cases that an alert can be linked to?
- Document this change in the "Overview" section of [View alert details](https://www.elastic.co/guide/en/security/master/alerts-ui-manage.html#view-alert-details). Make sure the updated screenshot shows that the alert has been attached to at least one case.

Issue #1813, opened 2022-04-05
[DOCS] Alert prevalence info added to alert details flyout  


nastasha-solomon:
## Description 

Alert prevalence data has been added to the alert details flyout. It expands on alert details and describes how the alert relates to other existing alerts. Alert prevalence data appears within the highlighted fields table as a new column. This data is affected by the time range selected for the Alerts table, but is not impacted by any other filters. 

<img width="641" alt="Screen Shot 2022-04-05 at 4 51 52 PM" src="https://user-images.githubusercontent.com/79124755/161848643-38f8cacb-41f1-4080-bb54-587be78b4654.png">

## Related dev issues
- https://github.com/elastic/security-team/issues/3055
- https://github.com/elastic/kibana/pull/127599
- https://github.com/elastic/kibana/pull/127474
- https://github.com/elastic/kibana/pull/128295

## Notes

- Need to update screenshot of alert detail flyout in the [View alert details](https://www.elastic.co/guide/en/security/master/alerts-ui-manage.html#view-alert-details) section and add some info about the "Alert prevalence" column. An example of how to read the data in the column might also be helpful.

PR #1825, opened 2022-04-11
[DOCS] Updates to alert details flyout 

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/1814 and https://github.com/elastic/security-docs/issues/1813

Preview here.


docs/detections/alerts-ui-manage.asciidoc
```diff
@@ -99,7 +99,11 @@ The alert summary section provides general alert details such as the alert's sta
...
 
-The *Highlighted Fields* section is also provided for all alerts and is customized to only show the most relevant fields for the alert type. Use this section to inform your triage efforts as you investigate the alert.
+Also described are the number of cases to which the alert has been added. An alert can be added to zero or more cases. Click case names to open case details.
+
+The *Highlighted Fields* section is also provided for all alerts and displays the most relevant fields for the alert type. Use this section to inform your triage efforts as you investigate the alert.
+
+Additional data about the alert's highlighted fields is given in the *Alert prevalence* column. This column shows the total number of other alerts within the selected timeframe that have identical values. For example, imagine an alert with an alert prevalence value of 3 for the `host.name` field. This means, within the given timeframe, three other alerts with the same `host.name` value exist. Alert prevalence data can help you investigate relationships with other alerts and gain more context about the event producing the alert.
```
joepeeples, line 106: This passage contains a lot of passive voice and discursive phrasing that could be streamlined.

* For the sake of conciseness, I don't think it's necessary to say how many cases an alert can be attached to, esp since in this context the user isn't trying to attach anything; they're just looking at info. Also adding something to "zero or more" is kind of impossible, so it could be omitted.
* The definition of what the prevalence column shows is already clear and straightforward, and the example seems kind of repetitive and wordy.

```suggestion
The alert details flyout also lists the number and names of cases to which the alert has been added. Click case names to open case details.

The *Highlighted Fields* section displays the most relevant fields for the alert type. Use this section to inform your triage efforts as you investigate the alert.

The *Alert prevalence* column shows the total number of other alerts within the selected timeframe that have identical values. Alert prevalence data can help you investigate relationships with other alerts and gain more context about the event producing the alert. 
```

michaelolo24 (approved):
Looks good, thanks! LGTM!

joepeeples (commented):
Added a few suggestions for streamlining the new content, thanks!

docs/detections/alerts-ui-manage.asciidoc
nastasha-solomon: Thanks for these edits, @joepeeples ! I debated adding the sentence about adding an alert because it didn't seem fully relevant to what was being described. It's also a bit misleading because the option to do that is in a sep menu. 

Thanks as well for your thoughts on the paragraph about the **Alert prevalence** column. I can see how the first sentence in the **Alert prevalence** section is unnecessary given the paragraph's proximity to the **Highlighted Fields** description--good call on removing that. I'm still torn on removing the example. I actually think it adds some context to the following sentence, which describes how to use alert prelavence data.

nastasha-solomon: Thanks for the reminder, @joepeeples ! Verified that the value in the prevalence column refers to the total number of alerts in the timeframe (so all other alerts and the current one). Will make that change in the next round of commits.

docs/detections/alerts-ui-manage.asciidoc
```diff
@@ -99,7 +99,11 @@ The alert summary section provides general alert details such as the alert's sta
 
 The alert's reason statement is under the alert summary. The reason statement describes what generated the alert and provides general alert details. You can use this information to understand the alert's origin and determine if the alert is relevant to your investigation.
 
-The *Highlighted Fields* section is also provided for all alerts and is customized to only show the most relevant fields for the alert type. Use this section to inform your triage efforts as you investigate the alert.
+The alert details flyout also lists the number and names of cases to which the alert has been added. Click case names to open case details.
```
benironside, line 102:
```suggestion
The alert details flyout also lists the number and names of cases to which the alert has been added. Click a case's name to view its details.
```
Could be wrong but if I understand correctly you would click on one case name.

benironside (commented):
LGTM, left one suggestion.

docs/detections/alerts-ui-manage.asciidoc
nastasha-solomon: You're correct and this makes sense - thanks @benironside ! I do want to stick with "open" over "view" though to avoid using (what could be construed as) ableist language.

karanverma-qasource:
Hi @nastasha-solomon ,

We tested the linked docs and found that all the docs are correctly updated as per the latest UI. So we are good to go ahead and merge the changes.

Marking this ticket as 'QA validated'.

Thanks!

jmikell821 (approved):
LGTM!

PR #1855, opened 2022-04-19
[DOCS] Update Action API examples

EvelienSchellekens:
Fix Action API examples of create and update a ServiceNow ticket.

`subActionParams` schema should consist of `incident` object and `comments`.


jmikell821:
Hi @EvelienSchellekens unfortunately, this PR brought in 400+ changed files and over 80 commits, so we had to reset the PR to the original commit - please check the **Commits** tab and ensure this is correct. I can take over the PR from here. Please let me know if this applies to `8.2` as well or any other previous version. Currently, `main` is equal to `8.3` until feature freeze.

EvelienSchellekens:
Hi @jmikell821, I checked the commit and it's correct. I think the changes also apply to all previous versions starting at 7.11 (that's when the incident object was introduced).

joepeeples (approved):
New structure for `incident` container LGTM!

PR #1908, opened 2022-04-26
[8.2] [DOCS] Field browser and runtime fields on Alerts table (55a23a85)

joepeeples:
NOTE: Due to a merge conflict mishap, the `main` version of this PR (#1894) accidentally got merged before everyone had a chance to review it. So we'll use this `8.2` version for review, and then "forward-port" any needed changes back to the `main` branch in a separate PR.

Backports the following commits to 8.2:
 - Merge branch 'issue-1834-field-browser-runtime-fields' (55a23a85)

Previews:
- [Manage detection alerts | Customize the Alerts table](https://security-docs_1908.docs-preview.app.elstc.co/guide/en/security/8.2/alerts-ui-manage.html#customize-the-alerts-table)
- [Add runtime fields to detection alerts](https://security-docs_1908.docs-preview.app.elstc.co/guide/en/security/8.2/alerts-runtime-fields.html) - Created new sub-topic page to avoid overextending "Manage detection alerts" topic

PR #1918, opened 2022-04-27
[main] [8.2] [DOCS] Field browser and runtime fields on Alerts table (55a23a85) (#1908)

joepeeples:
_Forward_-ports the following commits to main:
 - [8.2] [DOCS] Field browser and runtime fields on Alerts table (55a23a85) (#1908)


joepeeples:
No backport label needed, since this is actually a forward-port from `8.2`, so we don't need Mergify to create anything else.

Issue #1948, opened 2022-05-04
[BUG]Missing newly added template in 8.2.0 are missing under template doc


karanbirsingh-qasource:
## Description 
Missing newly added template in 8.2.0 are missing under template doc

## URL
https://www.elastic.co/guide/en/security/8.2/timeline-templates-ui.html#template-legend-ui

## Screen-Shoot

- **Actual Result**
![image](https://user-images.githubusercontent.com/59917825/166672012-0e3edbeb-e6f8-41d3-92f5-885bc1df881b.png)

- **Expected Result**

![image](https://user-images.githubusercontent.com/59917825/166671935-25cceeb6-d8e0-491b-a2b1-87ae430be8ff.png)

manishgupta-qasource:
Reviewed & assigned to @jmikell821

nastasha-solomon:
Chatted with @DefSecSentinel and will be adding the following summaries:

- **Comprehensive Process Timeline:** Useful for quickly assessing and investigating process related alerts.
- **Comprehensive File Timeline:** Useful for quickly assessing and investigating file related alerts.
- **Comprehensive Network Timeline:** Useful for quickly assessing and investigating network related alerts.
- **Comprehensive Registry Timeline:** Useful for quickly assessing and investigating registry related alerts.

karanbirsingh-qasource:
thanks for the update @nastasha-solomon  👍

PR #1961, opened 2022-05-04
[BUG]Missing newly added template in 8.2.0 are missing under template doc 

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/1948.

Preview [here](https://security-docs_1961.docs-preview.app.elstc.co/guide/en/security/master/timeline-templates-ui.html).


DefSecSentinel (approved):
LGTM

joepeeples (commented):
The new content is OK, but I think it'd be easier to read if the list was ordered semi-alphabetically or some other more logical way -- it's unclear how the items are currently ordered.

Also, the list is now a wall of text with a lot of repetition, without much unique or meaningful information about each timeline. It seems somewhat obvious from each timeline's name what kind of information and alerts it's related to, so we aren't really telling the reader much they didn't already know. Maybe we could trim down the descriptions and expand the lead-in paragraph to convey the same information with less repetition:

```
When you load {es-sec} prebuilt rules, some prebuilt Timeline templates are 
also loaded. You can attach these templates to detection rules. "Generic" 
templates use broad KQL queries to retrieve event data, and "comprehensive" 
templates use detailed KQL queries for additional information. The following 
templates are loaded:

* *Generic Endpoint Timeline*: Investigate Elastic Endpoint detection alerts.
* *Generic Network Timeline*: Investigate network-related detection alerts.
* *Generic Process Timeline*: Investigate process-related detection alerts.
* *Generic Threat Match Timeline*: Investigate threat indicator match detection alerts.
* *Comprehensive File Timeline*: Investigate file-related detection alerts.
* *Comprehensive Network Timeline*: Investigate network-related detection alerts.
* *Comprehensive Process Timeline*: Investigate process-related detection alerts. 
* *Comprehensive Registry Timeline*: Investigate registry-related detection alerts.
```

nastasha-solomon:
@joepeeples good point about the order. Another option I was considering was ordering them to align with the UI. That'd be the easiest to follow IMHO, but I'm open to either.

RE your suggestion to cut down the description: I've gone back and forth between using a short one-liner (like what was previously used) to explain the associated use case and being a bit more descriptive about the results users expect with the comprehensive and generic templates. For users who are familiar with Timeline, the difference between the two is probably obvious-- maybe even obvious to the point where they wouldn't be reading the docs. I'd expect the opposite for newer users and I think the longer description benefits them because it provides more information for each template. Sure, they could always open the templates in the Security app to compare their queries but wouldn't it be better for us to save them those extra clicks?

On a related note, these templates currently don't have descriptions in the UI. Maybe the shorter descriptions could be inserted there during `8.3`? @DefSecSentinel what do you think? 

Screen Shot 2022-05-06 at 3 51 39 PM

joepeeples:
> @joepeeples good point about the order. Another option I was considering was ordering them to align with the UI. That'd be the easiest to follow IMHO, but I'm open to either.

In this case the UI doesn't have a set order -- the templates are ordered by the **Last Modified** column, which is subject to change.

nastasha-solomon:
@joepeeples ah, I didn't notice that before. On second thought, maybe alphabetizing them is better since there is no set order. Plus, the generic templates might eventually be removed so lumping them together would make the removal process cleaner/easier whenever that change does happen.

docs/events/timeline-templates.asciidoc
```diff
@@ -23,15 +23,16 @@ alerts generated by the rule are investigated in Timeline, the alert's
...
-When you load {es-sec} prebuilt rules, some prebuilt Timeline templates are
-also loaded. You can attach these templates to detection rules. The following
-templates are loaded:
-
-* *Generic Endpoint Timeline*: Useful for investigating Elastic Endpoint
-security alerts.
-* *Generic Process Timeline*: Useful for investigating process-related alerts.
-* *Generic Network Timeline*: Useful for investigating network-related alerts.
-* *Generic Indicator Match Timeline*: Useful for investigating indicator match alerts.
+When you load {es-sec} prebuilt rules, some prebuilt Timeline templates are also loaded. You can attach these templates to detection rules. *Generic* templates use broad KQL queries to retrieve event data, and *Comprehensive* templates use detailed KQL queries for additional information. The following templates are loaded:
```
benironside, line 26:
```suggestion
When you load {es-sec} prebuilt rules, {es-sec} also loads a selection of prebuilt Timeline templates. You can attach these templates to detection rules. *Generic* templates use broad KQL queries to retrieve event data, and *Comprehensive* templates use detailed KQL queries to retrieve additional information. The following prebuilt templates appear by default:
```

benironside (commented):
Looks good! Left a couple of suggestions for your consideration.

docs/events/timeline-templates.asciidoc
```diff
@@ -23,15 +23,16 @@ alerts generated by the rule are investigated in Timeline, the alert's
...
-When you load {es-sec} prebuilt rules, some prebuilt Timeline templates are
-also loaded. You can attach these templates to detection rules. The following
-templates are loaded:
-
-* *Generic Endpoint Timeline*: Useful for investigating Elastic Endpoint
-security alerts.
-* *Generic Process Timeline*: Useful for investigating process-related alerts.
-* *Generic Network Timeline*: Useful for investigating network-related alerts.
-* *Generic Indicator Match Timeline*: Useful for investigating indicator match alerts.
+When you load {es-sec} prebuilt rules, {es-sec} also loads a selection of prebuilt Timeline templates. You can attach these templates to detection rules. *Generic* templates use broad KQL queries to retrieve event data, and *Comprehensive* templates use detailed KQL queries to retrieve additional information. The following prebuilt templates appear by default:
```
jmikell821, line 26:
```suggestion
When you load {elastic-sec} prebuilt rules, {elastic-sec} also loads a selection of prebuilt Timeline templates, which you can attach to detection rules. *Generic* templates use broad KQL queries to retrieve event data, and *Comprehensive* templates use detailed KQL queries to retrieve additional information. The following prebuilt templates appear by default:
```

docs/events/timeline-templates.asciidoc
```diff
@@ -23,15 +23,16 @@ alerts generated by the rule are investigated in Timeline, the alert's
...
+
+* *Generic Endpoint Timeline*: Investigate Elastic Endpoint detection alerts.
```
jmikell821, line 28:
```suggestion
* *Generic Endpoint Timeline*: Investigate {elastic-endpoint} detection alerts.
```

jmikell821 (commented):
Two small suggestions, then good to go. Thanks!

PR #2074, opened 2022-06-15
[DOCS] Revise workaround for aggregated fields in threshold rules

### LINKED ISSUE #2016: [DOCS] Threshold Rule actions can target aggregated fields directly
## Description 
We currently describe a workaround for targeting aggregated fields in Threshold Rule actions (https://github.com/elastic/security-docs/blame/349c716ee595144230ed51869f3d318c220462c1/docs/detections/rules-ui-create.asciidoc#L166-L178), however that workaround is no longer necessary. As of https://github.com/elastic/kibana/pull/94345 that step is no longer necessary, but we never removed the workaround from the docs.

We should remove this section from the docs, as users can more easily access the fields directly from the alert document, rather than use the `threshold_result` field copies.

## Acceptance Test Criteria
As a user, when configuring an action for a Threshold Rule which aggregates on (e.g.) `host.name`, I can directly reference the `host.name` field from a generated alert when configuring action templates for the rule.
 

## Notes

- Add all appropriate labels to the issue, especially the version number label. 
- Be sure to add any necessary screenshots, code text or console commands for clarity. 
- Include any conditions or caveats that may affect customers.

joepeeples:
Resolves #2016.

Preview: [Create a detection rule | Create a threshold rule](https://security-docs_2074.docs-preview.app.elstc.co/guide/en/security/master/rules-ui-create.html#create-threshold-rule)


madirey:
Thanks for this! Upon reading, I'm thinking we may want to leave some of this information... we could still point out that the generated signals/alerts are synthetic and don't contain all the fields (only those that appear in the Threshold field configuration of the rule). Something like: ```NOTE: Alerts created by *threshold* rules are synthetic alerts that do not resemble the source documents. The alert itself only contains data about the fields that were aggregated over (the `Group by` fields). Other fields are omitted, because they can vary across all source documents that were counted toward the threshold. Additionally, you can reference the actual count of documents that exceeded the threshold from the `kibana.alert.threshold_result.count` field`.```

joepeeples:
Thanks for the suggestion, @madirey! I added the note back in with your revisions; just let me know if anything else needs to be adjusted.

madirey (approved):
Thank you!!

jmikell821 (approved):
LGTM!

PR #2131, opened 2022-06-24
Add integrations 7.16.3 rule docs

terrancedejesus:
Security Doc updates for integrations rule changes in 8.2.

As discussed with @jmikell821, we are placing less in-depth review on these docs as a result of workload and timing. I will do an initial review to ensure all files exist related to what the site expects.


.gitignore
```diff
@@ -1,3 +1,4 @@
 .DS_Store
 docs/html_docs
 /html_docs
+launch.json
```
terrancedejesus, line 4: This was added during debugging and should be kept to ignore debug configurations.

terrancedejesus:
Changes look good.

Checks made:

- Ensured 7.16.3 folder was in `docs/detections/prebuilt-rules/downloadable-packages`
- Ensured `manual-updates.json` file does not exist in diff
- Ensured 7.16.3 strings exist in `docs/index.asciidoc`, `docs/detections/prebuilt-rules/prebuilt-rules-downloadable-updates.asciidoc` and `docs/detections/prebuilt-rules/prebuilt-rules-downloadable-updates.asciidoc`
- Reviewed labels are correct for this integrations

Issue #1923, opened 2022-04-29
[DOCS] Update Indicator Match rule documentation


nkhristinin:
## Description 

[Clients can have a problem](https://github.com/elastic/security-team/issues/3673) with the Indicator match rule and maxClauseCount.

So, if clients have an error with the IM rule like that:

`.. reason: "too_many_clauses: maxClauseCount is set to ...`

We should suggest client, in this case, to increase memory for Elasticsearch and maybe point them to [this document](https://www.elastic.co/blog/managing-and-troubleshooting-elasticsearch-memory)

Also, it's worth mentioning that for correct works of the IM rule you need to have **at least 2GB** of RAM

nastasha-solomon:
Merged #2162.

PR #2162, opened 2022-06-28
[DOCS] Update Indicator Match rule documentation

nastasha-solomon:
Addresses https://github.com/elastic/security-docs/issues/1923.

Preview here.

### DIFF
--- a/docs/troubleshooting/detections/detection-rules.asciidoc
+++ b/docs/troubleshooting/detections/detection-rules.asciidoc
@@ -51,6 +51,14 @@ If you receive the following rule failure: `"An error occurred during rule execu
 ====
 
 [discrete]
+[[IM-rule-heap-memory]]
+.Indicator match rules are failing due to insufficient heap memory
+[%collapsible]
+====
+If you receive the following rule failure: `Bulk Indexing of signals failed: index: ".index-name" reason: "maxClauseCount is set to 1024" type: "too_many_clauses"`, this indicates that the limit is too low for the total number of clauses that a query tree can have. Try increasing your cluster's heap memory.
+====
+
+[discrete]
 [[IM-slowness]]
 .General slowness
 [%collapsible]


docs/troubleshooting/detections/detection-rules.asciidoc
```diff
@@ -50,6 +50,14 @@ This can be caused by bad indicator data, a misconfigured rule, or too many even
 If you receive the following rule failure: `"An error occurred during rule execution: message: "Request Timeout after 90000ms"`, this indicates that the query phase is timing out. Try refining the time frame or dividing the data defined in the query into multiple rules.
 ====
 
+[discrete]
+[[IM-rule-heap-memory]]
+.Indicator match rules are failing due to insufficient heap memory
+[%collapsible]
+====
+If you receive the following rule failure: `Bulk Indexing of signals failed: index: ".index-name" reason: "maxClauseCount is set to 1024" type: "too_many_clauses"`, this indicates that the limit for the total number of clauses that a query tree can have is too low. Try increasing your cluster's heap memory to 2GB or more.
```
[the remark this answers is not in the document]
nastasha-solomon: @nkhristinin thanks for all the feedback!

I can revise the tip for increasing the ES JVM heap size and then point to the docs for doing that. Just want to verify that we're recommending that users set their heap size to 1GB and not 2GB (this was the amount I had taken down in my notes after researching the related dev issue and us meeting last week).

For the troubleshooting tip title, how about:
Indicator match rules are failing with a `maxClauseCount` error.

The version above doesn't describe the reason for the rules failing, but does indicate that the max clause count is involved. If 

Another revision, which is a bit more descriptive of the underlying cause: 
Indicator match rules are failing because the `maxClauseCount` limit is too low.

Let me know what you think - thanks!

nkhristinin (approved):
LGTM

docs/troubleshooting/detections/detection-rules.asciidoc
```diff
@@ -50,6 +50,14 @@ This can be caused by bad indicator data, a misconfigured rule, or too many even
 If you receive the following rule failure: `"An error occurred during rule execution: message: "Request Timeout after 90000ms"`, this indicates that the query phase is timing out. Try refining the time frame or dividing the data defined in the query into multiple rules.
 ====
 
+[discrete]
+[[IM-rule-heap-memory]]
+.Indicator match rules are failing because the `maxClauseCount` limit is too low.
```
joepeeples, line 55: Remove period to match other headings
```suggestion
.Indicator match rules are failing because the `maxClauseCount` limit is too low
```

docs/troubleshooting/detections/detection-rules.asciidoc
```diff
@@ -50,6 +50,14 @@ This can be caused by bad indicator data, a misconfigured rule, or too many even
...
+[%collapsible]
+====
+If you receive the following rule failure: `Bulk Indexing of signals failed: index: ".index-name" reason: "maxClauseCount is set to 1024" type: "too_many_clauses"`, this indicates that the limit for the total number of clauses that a query tree can have is too low. To update your maximum clause count, {ref}/advanced-configuration.html#set-jvm-heap-size[increase the size of your {es} JVM heap memory]. 1GB of {es} JVM heap size or more is sufficient.
```
joepeeples, line 58: I noticed that the linked page in ES docs doesn't use a space for measurements, but I think including a space is much more standard in general. 
```suggestion
If you receive the following rule failure: `Bulk Indexing of signals failed: index: ".index-name" reason: "maxClauseCount is set to 1024" type: "too_many_clauses"`, this indicates that the limit for the total number of clauses that a query tree can have is too low. To update your maximum clause count, {ref}/advanced-configuration.html#set-jvm-heap-size[increase the size of your {es} JVM heap memory]. 1 GB of {es} JVM heap size or more is sufficient.
```

joepeeples (commented):
Two tiny edits, otherwise LGTM!

docs/troubleshooting/detections/detection-rules.asciidoc
nastasha-solomon: Either work IMO and I've seen both in our solution docs. :)

benironside (approved):
LGTM!

jmikell821 (approved):

Issue #1857, opened 2022-04-19
[BUG] Refresh section about refining Timeline results


nastasha-solomon:
Docs about [refining the Timeline results](https://www.elastic.co/guide/en/security/master/timelines-ui.html#refine-timeline-results) need to be refreshed. The option to do this is currently located in the **Data view** menu. 

Selecting the **Show only detection alerts** option shows alerts only. If it's not selected, both events and alerts are shown. Default state is de-selected.

![data view](https://user-images.githubusercontent.com/79124755/164084958-50f23cd6-de90-4506-a377-6795181eb337.png)

PR #2204, opened 2022-07-18
[BUG] Refresh section about refining Timeline results

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/1857.

Preview [here](https://security-docs_2204.docs-preview.app.elstc.co/guide/en/security/master/timelines-ui.html#refine-timeline-results).


docs/events/timeline-ui-overview.asciidoc
```diff
@@ -30,8 +30,7 @@ retrieved from the alert. For more information, refer to <<timeline-templates-ui
 [[refine-timeline-results]]
 == View and refine Timeline results
 
-Using the drop-down options by the KQL bar, you can select whether <<det-engine-terminology, alerts>>,
-other raw events, or both are displayed in the Timeline.
+Go to the right of the date and time picker, then click *Data view*. To hide raw events and display <<det-engine-terminology, detection alerts>> only, select *Show only detection alerts*. This setting is not selected by default. Keep it de-selected to display both raw events and detection alerts in Timeline.
```
joepeeples, line 33: - The previous version had a nice summary of what exactly this setting does, so it might help to lead with something similar. Otherwise this just starts with instructions that the user might not know why they should follow.
- No need to explain that the user has to move their mouse to the right of the time picker; just say the drop-down is to the right.
- It's a little confusing to first tell the user "select this checkbox" and then in the same paragraph say "keep this deselected." We can tell them what the checkbox does when they select it, and it's pretty clear what happens when it's not selected.
```suggestion
You can select whether Timeline displays <<det-engine-terminology, detection alerts>> and other raw events, or just alerts. By default, Timeline displays both raw events and alerts. To hide raw events and display alerts only, click *Data view* to the right of the date and time picker, then select *Show only detection alerts*.
```

joepeeples (commented):
Added some thoughts and suggestions, thanks!

docs/events/timeline-ui-overview.asciidoc
nastasha-solomon: @joepeeples all good points and I can get behind these changes. I figured the section title provided enough lead-in but I can see how setting the stage a bit more (like you did) creates a friendlier tone. I also like the way you ordered the information in your revision. It makes sense to start off with a general statement and then described the default state/setting and how to change it.

benironside (approved):
lgtm!

jmikell821 (approved):
LGTM 👍

michaelolo24 (approved):
Thanks! LGTM!

Issue #2260, opened 2022-08-04
[DOCS][CSPM][8.4] Add KSPM docs for 8.4


benironside:
## Description 

the CSPM team is introducing KSPM capabilities in 8.4. These include a new integration ("Kubernetes Security Posture Management"), a new dashboard ("Cloud posture"), a new top-level page called "Findings", and a new page under Manage called "Benchmarks". This issue is to document the new pages, the new integration, and how to deploy the daemonset that collects kubernetes data to populate the new pages.

The documentation plan / outline is [here](https://docs.google.com/document/d/1gqRAz3hDkJUKiHl4DCFd-CIVyyViUFeZ19GrVC3EZx8/edit)

## Acceptance Test Criteria

Docs PRs have Tinsae's sign-off, and maybe sign-off from other folks in the Cloud Security team. 

## Sub-issues
#2271 
#2272 
#2273 
#2274

PR #2289, opened 2022-08-10
[DOCS] Response console UI for endpoint actions

joepeeples:
Resolves #2211.

Previews:
- [Endpoints](https://security-docs_2289.docs-preview.app.elstc.co/guide/en/security/master/admin-page-ov.html) — Minor cleanup, revised to link to new topic
- [Endpoint response actions](https://security-docs_2289.docs-preview.app.elstc.co/guide/en/security/master/response-actions.html) — New topic
- [Host isolation](https://security-docs_2289.docs-preview.app.elstc.co/guide/en/security/master/host-isolation-ov.html) — Revised & restructured to include response console
- [Troubleshooting | Endpoint management](https://security-docs_2289.docs-preview.app.elstc.co/guide/en/security/master/ts-management.html) — Minor edit, renamed topic
- [Manage detection alerts | Take action on an alert](https://security-docs_2289.docs-preview.app.elstc.co/guide/en/security/master/alerts-ui-manage.html#alert-actions) - Added link to new topic

PR #2315, opened 2022-08-17
[DOCS] Creates stubs for new KSPM pages

benironside:
Addresses #2260 by adding stub pages for the new KSPM docs to address broken link/build failure issues.


joepeeples:
I think the links are OK, but the page locations might need a little reorganizing. As of https://github.com/elastic/security-docs/pull/2289 we're renaming the "Manage" section to "Endpoint management," so that the overall section is no longer mirroring what's in the nav but just describing a general function around endpoints.

It'd probably be better to have the "Benchmark rules" page in a separate main section that's all about cloud stuff, which I don't think we currently have but could create. So instead of adding that that file as an include in `manage-intro.asciidoc`, you could add it to the index/intro file for a cloud section. (Would that cloud index be `cloud-native-security-overview.asciidoc`? Where is _that_ page going to be located in the docs?)

benironside:
> I think the links are OK, but the page locations might need a little reorganizing. As of #2289 we're renaming the "Manage" section to "Endpoint management," so that the overall section is no longer mirroring what's in the nav but just describing a general function around endpoints.
> 
> It'd probably be better to have the "Benchmark rules" page in a separate main section that's all about cloud stuff, which I don't think we currently have but could create. So instead of adding that file as an include in `manage-intro.asciidoc`, you could add it to the index/intro file for a cloud section. (Would that be `cloud-native-security-overview.asciidoc`? Where is _that_ page going to be located in the docs?)

Sounds good to add the Benchmark rules page to the Cloud Native Security section, thanks. That section is going to go under Detections and Alerts.

docs/cloud-native-security/cloud-native-security-index.asciidoc
```diff
@@ -0,0 +1,7 @@
+[[cloud-native-security-overview]]
+
+= Cloud native security
+Elastic’s cloud security capabilities help you to improve your Kubernetes security posture by comparing your configuration to best practices, and help you monitor and investigate your Linux deployments inside and outside of Kubernetes. This is an area of focus for Elastic Security, and you can expect many more cloud security features in future releases.
```
joepeeples, line 4: The second sentence sounds more like a blog or marketing statement, and even though it's pretty general we try to avoid forward-looking statements that seem to "promise" things.

I know the focus of this PR is more about structure than content, so we can do a closer review later if there's more to come for this section.

joepeeples (commented):
Structure looks good! Exciting to see new cloud features getting top billing in the docs.

docs/cloud-native-security/cloud-native-security-index.asciidoc
jmikell821: I would definitely nix the second sentence.

jmikell821 (commented):
@benironside just one small comment piggybacking off what @joepeeples mentioned, otherwise LGTM!

PR #2321, opened 2022-08-18
Update index-api-overview.asciidoc

jeanfabrice:
Add a note to suggest the user to update the Signal ILM policy with a delete action


jmikell821:
@elasticmachine run elasticsearch-ci/docs

docs/detections/api/rules/index-api-overview.asciidoc
```diff
@@ -33,6 +33,8 @@ policy is created for the signal index:
 
 The `policy` and `rollover_alias` use the same name as the signal index.
 
+NOTE: It is highly recommended to update this ILM policy to include an {ref}/ilm-delete.html[{ilm-init} delete action], based on your internal security compliance policy, if you don't want the signal indices to stay forever on your deployment's hot tiers
```
lockewritesdocs, line 36:
```suggestion
NOTE: To reduce clutter on your hot tier, we highly recommend adding a {ref}/ilm-delete.html[delete action] to this {ilm-init} policy. Otherwise, the signal indices will remain on your hot tier indefinitely. 
```

lockewritesdocs: I suggest moving the gain (to reduce clutter) first, and keeping the message more succinct. If including the bit about the internal security compliance policy is needed, you could say something like:
>NOTE: To reduce clutter on your hot tier, we highly recommend adding a {ref}/ilm-delete.html[delete action] to this {ilm-init} policy that complies with your internal security requirements. Otherwise, the signal indices will remain on your hot tier indefinitely.

lockewritesdocs (commented):
Left a suggested change, plus an alternative wording.

docs/detections/api/rules/index-api-overview.asciidoc
jeanfabrice: love the alternative <3

lockewritesdocs (approved):
LGTM 🦖

Issue #2304, opened 2022-08-12
[DOCS][Security Solution][Exceptions] - New exceptions UI/UX


yctercero:
## Description
We are updating the exceptions UI/UX. Much of it is not necessarily new functionality, but updated flows. To be a bit more specific, a lot of this already exists on the API for exceptions, but was not exposed in the UI. However, it is going to touch a lot of surface area.

Ticket will be updated as we merge changes in.

[Exceptions Epic](https://github.com/elastic/security-team/issues/2089)

[Update 9.14.22] - much of epic has been pushed to 8.6. This ticket will reflect only what is going in 8.5.

### Rule details exceptions tab

- Exceptions tab is now split into `Rule exceptions` tab and `Endpoint exceptions` tab [[PR](https://github.com/elastic/kibana/pull/138770)]
    - If the rule is not linked to the endpoint exception list, the `Endpoint exception` tab will not be shown 
- The exception item UI has been updated [[PR](https://github.com/elastic/kibana/pull/138770)]
    - It now includes an `Affects x rules` button that shows the user which rules are affected by the exception item
        - Clicking on it shows the rules and user can click to navigate to that rule
- The view for when no exception items exists is updated [[PR](https://github.com/elastic/kibana/pull/138770)]

The exceptions list management page, edit/add modal all remain unchanged.

yctercero:
Release notes for PR:

- https://github.com/elastic/kibana/pull/138770 - Separates exceptions into different tabs - "Rule exceptions" and "Endpoint exceptions". Updates the exception item UI to include information on how many rules are affected by said exception item.
- https://github.com/elastic/kibana/pull/141682 (not sure if this will be added to release notes, I put release_note:skip because it's just minor bug fixes to the PR above). Screenshots are added in this PR - not sure if you need to wait on this PR to go in before updating any exception screenshots in docs.

nastasha-solomon:
@yctercero thanks for providing summaries for both! If the bugs fixed in https://github.com/elastic/kibana/pull/141682 were discovered internally (i.e., they weren’t reported by one of our customers), it’s likely fine to omit them from the 8.5.0 release notes.

yctercero:
> @yctercero thanks for providing summaries for both! If the bugs fixed in [elastic/kibana#141682](https://github.com/elastic/kibana/pull/141682) were discovered internally (i.e., they weren’t reported by one of our customers), it’s likely fine to omit them from the 8.5.0 release notes.

[...]

PR #2546, opened 2022-10-05
[DOCS] New exceptions UI/UX

nastasha-solomon:
Addresses https://github.com/elastic/security-docs/issues/2304, which includes the following changes:
- New functionality that shows users how many rules an exception affects. I added a line to the start of the topic indicating that an exception could affect multiple rules. I also created a short sub-section for the feature titled "Find rules using the same exception". Feel free to suggest a different title.
- The exceptions tab was split into two tabs: the **Rule exceptions** tab and the **Endpoint exceptions** tab. A few notes here:
   - The docs that explained how to add an endpoint exception in 8.4 and earlier already explained the pre-reqs/conditions needed to create an endpoint exception so I didn't change any of it.
   - I changed the screenshot that showed the rule exception tab. Previously, it showed the Alerts tab with the trend table and highlighted the exceptions tab. My screenshot update is the more literal representation :))

Previews:
- [Rule exceptions and value lists](https://security-docs_2546.docs-preview.app.elstc.co/guide/en/security/master/detections-ui-exceptions.html)
- [Add exceptions to a rule](https://security-docs_2546.docs-preview.app.elstc.co/guide/en/security/master/detections-ui-exceptions.html#endpoint-rule-exceptions) - Updated step 1 and the screenshot
- [Add Elastic endpoint exceptions](https://security-docs_2546.docs-preview.app.elstc.co/guide/en/security/master/detections-ui-exceptions.html#endpoint-rule-exceptions) - Updated first step
- [Find rules using the same exception](https://security-docs_2546.docs-preview.app.elstc.co/guide/en/security/master/detections-ui-exceptions.html#rules-using-same-exception) - New section


docs/detections/detections-ui-exceptions.asciidoc
```diff
@@ -6,7 +6,7 @@ To prevent the creation of unwanted alerts, you can add exceptions to any rule.
 alerts are not generated. They provide a convenient way of allowing trusted
 processes and network activity to function without producing unnecessary noise.
 
-You can add multiple exceptions to one rule.
+You can add multiple exceptions to a single rule. An exception can also affect multiple rules.
```
benironside, line 9:
```suggestion
You can add multiple exceptions to a single rule. An exception can also apply to multiple rules.
```
if you think this makes sense

docs/detections/detections-ui-exceptions.asciidoc
```diff
@@ -278,6 +274,14 @@ Creates an exception that excludes all LFC-signed trusted processes:
 [role="screenshot"]
 image::images/nested-exp.png[]
 
+[float]
+[[rules-using-same-exception]]
+=== Find rules using the same exception
+From the *Rule exceptions* or *Endpoint exceptions* tab, navigate to an exception list item and click *Affects X rules*. The rule or rules that the exception was added to displays.
```
benironside, line 280:
```suggestion
From the *Rule exceptions* or *Endpoint exceptions* tab, navigate to an exception list item and click *Affects X rules* to display the rules affected by the exception.
```

benironside (commented):
Looks great, left a couple minor suggestions.

yctercero (approved):
Looks awesome! Thanks for the updates.

docs/detections/detections-ui-exceptions.asciidoc
```diff
@@ -102,11 +102,10 @@ specific event in the sequence, update the rule's EQL statement. For example:
 * To add an exception from the rule details page:
 .. Go to the rule details page of the rule to which you want to add an
 exception (*Manage* -> *Rules* -> *_<Rule name>_*).
-.. Scroll down below the rule details and select the *Exceptions* tab.
+.. Scroll down the rule details page, select the *Rule Exceptions* tab, then click *Add rule exception*.
```
joepeeples, line 105:
```suggestion
.. Scroll down the rule details page, select the *Rule exceptions* tab, then click *Add rule exception*.
```

docs/detections/detections-ui-exceptions.asciidoc
```diff
@@ -278,6 +274,14 @@ Creates an exception that excludes all LFC-signed trusted processes:
 [role="screenshot"]
 image::images/nested-exp.png[]
 
+[float]
+[[rules-using-same-exception]]
+=== Find rules using the same exception
+From the *Rule exceptions* or *Endpoint exceptions* tab, navigate to an exception list item and click *Affects X rules* to display the rules affected by the exception.
```
joepeeples, line 280: This section could use a little more context, maybe an extra sentence easing the reader into the idea of related/affected rules before jumping into actual instructions. Something like "If an exception is used by other rules, this is indicated on the exception" (not really an elegant sentence, but that general idea).

joepeeples (commented):
Looks good! One edit and a suggestion, thanks!

docs/detections/detections-ui-exceptions.asciidoc
nastasha-solomon: How about:

```suggestion
An exception might be used by other rules. To find out, click on the *Rule exceptions* or *Endpoint exceptions* tab, navigate to an exception list item, then click *Affects X rules*. 
```

jmikell821: This is good -- I like `select` vs. `click on`, especially when the user has a choice. I also think `X` is supposed to be in italics, (_x_), but that may not render with the bold already there -- can't remember. It may be OK. 

You could also flip this if you wanted: 

`"To find out if an exception is used by other rules...." `

jmikell821 (commented):
One slight comment - LGTM otherwise 👍

docs/detections/detections-ui-exceptions.asciidoc
nastasha-solomon:
```suggestion
To find out if an exception is used by other rules, select the *Rule exceptions* or *Endpoint exceptions* tab, navigate to an exception list item, then click *Affects _X_ rules*. 
```

PR #2566, opened 2022-10-12
[DOCS] Add Terminal output view information to Session view doc

### LINKED ISSUE #2458: [DOCS] Terminal Output Review for Session View
For the 8.5 Release, we will release Terminal Output Review for Session View
https://github.com/elastic/security-team/issues/2644
[after reading the ticket]

## Description
This new feature will allow security analysts to view the terminal output in the session viewer to understand a compromised session better.
The ticket mentioned above has screenshots and video recordings of how navigation works and the information displayed.

## What's needed
- Outline the business benefit from infosec analyst
- Outline how to enable output within settings; current Terminal Output has a default = off
- Outlined the workflow between session view jumping in and out of the terminal view
- Outline the controls within the terminal view
- Outline the index where to expect terminal output data streams

benironside:
Fixes #2458 by updating the Session view doc with information about Terminal output view. 

AWP team, please review for accuracy and completeness. Thanks!

Preview: [Session view](https://security-docs_2566.docs-preview.app.elstc.co/guide/en/security/master/session-view.html)


docs/detections/session-view.asciidoc
```diff
@@ -16,20 +16,23 @@ and investigating session activity on your Linux infrastructure and understandin
...
+* *TTY output:* Terminal output associated with each process in the session.
 
 NOTE: To view Linux session data from your Kubernetes infrastructure, you'll need to set up the <<kubernetes-dashboard,Kubernetes dashboard>>.
 
 [float]
 [[enable-session-view]]
 === Enable Session View data
 Session View uses process data collected by the {elastic-defend} integration,
-but this data is not collected by default. To enable Session View data, go to *Manage* -> *Policies*
+but this data is not collected by default. To enable Session View data: go to *Manage* -> *Policies*
```
opauloh, line 27: Starting from 8.5 Session data is enabled by default when the user selects Cloud Workload integrations.

image


But the steps are still valuable for enabling it in Endpoint integrations and to enable TTY output

opauloh:
Hi @benironside 

Here is some additional info regarding the TTY, if you find it relevant to add:

- Hovering the markers displays the process name and clicking on the marker also navigates to the process output.
- Playback controls: Click anywhere in the progress bar to navigate 

Also, saw some things regarding the session view that changed in 8.5 (which can be addressed in a separate PR as well):

- Session View is no longer Beta, but Terminal Output is. Should we move the Beta warning at the beginning to the Terminal Output section?
- Should we include that Session View is part of a [subscription](https://www.elastic.co/subscriptions) (Enterprise) plan? (Workload session auditing)

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +89,23 @@ Session View includes an additional badge not pictured above:
 * The *Exec user change* badge highlights exec user changes, such as when a user escalates to root:
 +
 [role="screenshot"]
-image::images/session-view-exec-user-change-badge.png[The Exec user change button,width=80%,height=80%]
+image::images/session-view-exec-user-change-badge.png[The Exec user change badge,width=80%,height=80%]
+
+* The *Output* badge appears next to commands which generated TTY output. Click it to open Terminal output view.
++
+[role="screenshot"]
+image::images/session-view-output-badge.png[The Output badge,width=80%,height=30%]
```
mitodrummer, line 97: nit pick: a screen shot of a command that isn't a typo might look better

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +89,23 @@ Session View includes an additional badge not pictured above:
...
+
+[[session-view-output]]
+[discrete]
+=== Terminal output view UI
+Terminal output view can help provide a more complete understanding of commands executed by users or adversaries, and assist with auditing and compliance objectives. It allows you to review a session's TTY output, in the context of the commands which generated it. It includes both TTY output, and commands entered into a shell as typed — provided the shell was in echo mode (for background information about Linux terminals, refer to https://github.com/elastic/security-team/blob/main/docs/adaptive-workload-protection-team/architecture/linux-terminals.md[Linux terminals]).
```
mitodrummer, line 102: security-team repo is private, @qcorporation @m-sample do we know of a good place to put this material for public viewing?

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +89,23 @@ Session View includes an additional badge not pictured above:
...
+
+[role="screenshot"]
+image::images/session-view-output-viewer.png[Terminal output view,width=100%,height=100%]
+
+1. The search bar. Use it to find and highlight search terms within the current session.
+The buttons on its right side allow you to jump through search results.
+2. The right-side scroll bar. Use it along with the bottom-side scroll bar to navigate output data that doesn't fit on a single screen.
+3. The playback controls: *Start*, *Prev*, *Play*, *Next*, and *Last*. Use them to advance or rewind the session's commands and output.
```
mitodrummer, line 110: Prev and Next buttons will move the playhead between the processes generating output. kind of like skipping tracks on a CD

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +89,23 @@ Session View includes an additional badge not pictured above:
...
+4. The *Fit screen*, *Zoom in*, and *Zoom out* buttons. Use them to adjust the text size.
```
mitodrummer, line 111: might be worth noting that using the existing "fullscreen" toggle is advised when viewing sessions whose host window size is large. In order to accurately render terminal output for graphical programs like vim, the sessions tty row and column configuration is captured. You can see this displayed in the top right corner of the terminal window boundary. A users terminal can also be resized at any point during a session, so this will be reflected when playing back terminal output.

mitodrummer (changes requested):
Looking good, some notes, but also, we probably need a section to describe some of the advanced configuration options that are available.

image
image
image

These settings are found at the bottom of the edit integration policy page.

**max_kilobytes_per_process** basically tells endpoint when to stop recording output for an instance of a process.

**max_kilobytes_per_event** controls how big an event gets before it gets sent to elastic and a new event continues recording for the process. basically our "batch size". 

**max_event_interval_seconds** how long to wait for the next tty_write call before sending the event to elastic. As an example, if you started a session, and typed nothing and then waited. The endpoint would wait 30 seconds (since bash hasn't exited yet), to send up the output generated so far (most likely a welcome message and an empty prompt). This setting could be reduced to say 5 seconds, if a user cares about seeing more frequent updates. 

It's worth mentioning that we only record a single timestamp per batch/event, so if a user requires more granular timestamps they can reduce both max_kilobytes_per_event and max_event_interval_seconds to achieve this. Higher values are recommend to save on storage and network performance.

docs/detections/session-view.asciidoc
benironside: 100% lol, my brain fixed it 🙄

benironside:
Thank you @mitodrummer and @opauloh for the great feedback. Incorporated it.

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +89,34 @@ Session View includes an additional badge not pictured above:
...
+* Select the maximum duration in seconds for which output is batched by modifying `linux.advanced.tty_io.max_event_interval_seconds`. Defaults to 30 seconds. For example, with the default value, any recorded output will be sent to {es} every 30 seconds (unless it first exceeds the `max_kilobytes_per_event` value, in which case it might be sent sooner).
+
+[role="screenshot"]
+image::images/session-view-output-viewer.png[Terminal output view,width=100%,height=100%]
+
+1. The search bar. Use to find and highlight search terms within the current session.
+The buttons on its right side allow you to jump through search results.
+2. The right-side scroll bar. Use along with the bottom-side scroll bar to navigate output data that doesn't fit on a single screen.
+3. The playback controls: *Start*, *Prev*, *Play*, *Next*, *Last*, and the progress bar. Use to advance or rewind the session's commands and output. Click anywhere on the progress bar to jump to that part of the session. The marks on the timeline represent processes that generated output, click them or click the *Prev* and *Next* buttons to skip between processes.
```
benironside, line 121:
```suggestion
3. The playback controls: *Start*, *Prev*, *Play*, *Next*, *Last*, and the progress bar. Use to advance or rewind the session's commands and output. Click anywhere on the progress bar to jump to that part of the session. The marks on the bar represent processes that generated output, click them or click the *Prev* and *Next* buttons to skip between processes.
```

docs/detections/session-view.asciidoc
qcorporation: Yeah, security-team and docs.elastic.co are internal resources. If we want to include this as a public-facing doc then we're going to have to pull in linux-terminals.md into the docs repo
@benironside is your intention to making linux-terminals.md public facing? If so then we'll need to pull that out of the private repo

benironside: @qcorporation I'd like to be able to link to it, IF the team feels it's appropriate to make it public facing. The background info seems relevant here.

m-sample: > @m-sample can we make this public? (is it ready?) https://github.com/elastic/security-team/blob/main/docs/adaptive-workload-protection-team/architecture/linux-terminals.md

It will need a clean up to make it broadly consumable.  It is also very terse and assumes the reader has a very good understanding of Linux (e.g. what is a process group, signals etc).  That said I think we *should* share some technical outline of how TTY's work to help customers really understand this feature.

benironside: Unless anyone objects and suggests another resource, I'll remove it for now, and create a docs issue for adding a link with background info there.

qcorporation (approved):

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,35 @@ Session View includes an additional badge not pictured above:
...
+* Select the maximum number of kilobytes of output to send to {es} as a single event by modifying `linux.advanced.tty_io.max_kilobytes_per_event`. Defaults to 512 KB. For example, with the default value, the maximum amount of output that will be sent to {es} as a single event is 512 KB. Additional data will be captured as a new event.
+* Select the maximum duration in seconds for which output is batched by modifying `linux.advanced.tty_io.max_event_interval_seconds`. Defaults to 30 seconds. For example, with the default value, any recorded output will be sent to {es} every 30 seconds (unless it first exceeds the `max_kilobytes_per_event` value, in which case it might be sent sooner).
+
+[role="screenshot"]
+image::images/session-view-output-viewer.png[Terminal output view,width=100%,height=100%]
+
+1. The search bar. Use to find and highlight search terms within the current session.
+The buttons on its right side allow you to jump through search results.
+2. The right-side scroll bar. Use along with the bottom-side scroll bar to navigate output data that doesn't fit on a single screen.
```
jmikell821, line 125:
```suggestion
2. The right-side scroll bar. Use along with the bottom scroll bar to navigate output data that doesn't fit on a single screen.
```

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,35 @@ Session View includes an additional badge not pictured above:
...
+3. The playback controls: *Start*, *Prev*, *Play*, *Next*, *Last*, and the progress bar. Use to advance or rewind the session's commands and output. Click anywhere on the progress bar to jump to that part of the session. The marks on the bar represent processes that generated output, click them or click the *Prev* and *Next* buttons to skip between processes.
```
jmikell821, line 126:
```suggestion
3. The playback controls: *Start*, *Prev*, *Play*, *Next*, *Last*, and the progress bar. Use to advance or rewind the session's commands and output. Click anywhere on the progress bar to jump to that part of the session. The marks on the bar represent processes that generated output. Click them or the *Prev* and *Next* buttons to skip between processes.
```

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,35 @@ Session View includes an additional badge not pictured above:
...
+[role="screenshot"]
+image::images/session-view-output-badge.png[The Output badge,width=80%,height=80%]
+
+[[session-view-output]]
+[discrete]
+=== Terminal output view UI
+
+beta::[]
+
+Terminal output view helps provide a more complete understanding of commands executed by users or adversaries, and assists with auditing and compliance objectives. It allows you to review a session's TTY output, in the context of the commands which generated it. It includes both TTY output, and commands entered into a shell as typed — provided the shell was in echo mode.
```
jmikell821, line 109: General suggestion -- feel free to tweak here if I messed up the context. 
```suggestion
Terminal output view helps provide a more complete understanding of commands executed by users or adversaries, and assists with auditing and compliance objectives. It allows you to review a session's TTY output in the context of the commands which generated it. It includes both TTY output, and shell commands entered as  `—`, if the shell was in echo mode.
```

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,35 @@ Session View includes an additional badge not pictured above:
...
+
+To enable Terminal output data capture: go to *Manage* -> *Policies*
+and edit one or more of your {elastic-defend} integration policies. On the *Policy settings* tab,
+scroll down to the Linux event collection section near the bottom of the page
+and turn on *Include session data* and *Capture terminal output*. You can configure several additional settings by clicking *Advanced settings* at the bottom of the page:
```
jmikell821, line 114: These are steps, so make this an ordered list.
```suggestion
To enable Terminal output data capture: 
. Go to *Manage* -> *Policies*, then select one or more of your {elastic-defend} integration policies to edit. 
. On the *Policy settings* tab, scroll down to the Linux event collection section near the bottom of the page
and select the *Include session data* and *Capture terminal output* options. 

You can configure several additional settings by clicking *Advanced settings* at the bottom of the page:
```

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,35 @@ Session View includes an additional badge not pictured above:
...
+
+[role="screenshot"]
+image::images/session-view-output-viewer.png[Terminal output view,width=100%,height=100%]
+
+1. The search bar. Use to find and highlight search terms within the current session.
+The buttons on its right side allow you to jump through search results.
+2. The right-side scroll bar. Use along with the bottom-side scroll bar to navigate output data that doesn't fit on a single screen.
+3. The playback controls: *Start*, *Prev*, *Play*, *Next*, *Last*, and the progress bar. Use to advance or rewind the session's commands and output. Click anywhere on the progress bar to jump to that part of the session. The marks on the bar represent processes that generated output, click them or click the *Prev* and *Next* buttons to skip between processes.
+4. The *Fit screen*, *Zoom in*, and *Zoom out* buttons. Use to adjust the text size. Use Session view's *Fullscreen* button (located next to the *Close session viewer* button) to improve the display of output that doesn't easily fit, especially for graphical programs like `vim`.
```
jmikell821, line 127: I didn't think the last sentence was necessary, but feel free to veto me on this one. 
```suggestion
4. The *Fit screen*, *Zoom in*, and *Zoom out* buttons. Use to adjust the text size. Use Session view's *Fullscreen* button (located next to the *Close session viewer* button) to expand the display output.
```

jmikell821 (commented):
@benironside left some general suggestions for conciseness and consistency. In some places, you have "`session view button`" and others, "`session view icon`".  I'm trying to find our UX ponderisms page, but based on our [manage detection alerts topic](https://www.elastic.co/guide/en/security/current/alerts-ui-manage.html), most of the table elements are listed as "buttons." Please make sure all those references are the same. 

I'd like us all to review this again with fresh eyes once feedback is merged. Thanks!

docs/detections/session-view.asciidoc
```diff
@@ -16,20 +16,27 @@ and investigating session activity on your Linux infrastructure and understandin
...
-collected by {elastic-defend} when this setting was enabled. For more information about the additional
+but this data is not always collected by default. To confirm that Session View data is enabled:
+
+* Go to *Manage* -> *Policies*, and edit one or more of your {elastic-defend} integration policies.
+* Select the *Policy settings* tab, and
+scroll down to the Linux event collection section near the bottom.
+Turn on the *Include session data* toggle, if it isn't on already.
+* To enable TTY output capture, turn on the *Capture terminal output* toggle.
+
+Session View can only display data that was collected by {elastic-defend} when *Include session data* was enabled. For more information about the additional
```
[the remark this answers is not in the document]
benironside: Disagree here. If you turn the setting off (it "is" off), you can still see any historical data collected when it "was" on.

docs/detections/session-view.asciidoc
```diff
@@ -1,13 +1,13 @@
...
 == Session View
 
-beta::[]
-
 Session View is an investigation tool that allows you to examine Linux process data organized
 in a tree-like structure according to the Linux logical event model, with processes organized by parentage and time of execution.
 It displays events in a highly readable format that is inspired by the terminal. This makes it a powerful tool for monitoring
 and investigating session activity on your Linux infrastructure and understanding user and service behavior.
 
+NOTE: Session View requires an www.elastic.co/pricing[Enterprise subscription].
```
joepeeples, line 9: Reformat as a Requirements sidebar?

docs/detections/session-view.asciidoc
```diff
@@ -16,20 +16,27 @@ and investigating session activity on your Linux infrastructure and understandin
...
-scroll down to the Linux event collection section near the bottom of the page
-and turn on the *Include session data* toggle. Session View can only display data that was
-collected by {elastic-defend} when this setting was enabled. For more information about the additional
+but this data is not always collected by default. To confirm that Session View data is enabled:
+
+* Go to *Manage* -> *Policies*, and edit one or more of your {elastic-defend} integration policies.
+* Select the *Policy settings* tab, then
+scroll down to the Linux event collection section near the bottom.
+Turn on the *Include session data* toggle if it isn't on already.
+* To enable TTY output capture, turn on the *Capture terminal output* toggle.
```
joepeeples, line 33: Change list to numbered steps, since these are done in a specific order

docs/detections/session-view.asciidoc
```diff
@@ -40,11 +47,11 @@ Events and sessions that you can investigate in Session View have a rectangular
...
 Look for alerts that have the **Open Session View** button in the **Actions** column:
 [role="screenshot"]
-image::images/session-view-action-icon-detail.png[Detail of the Open Session View icon,width=75%]
+image::images/session-view-action-icon-detail.png[Detail of the Open Session View button,width=75%]
 
 * On the Hosts page (*Explore* -> *Hosts*), select the *Sessions* or the *Events* tab.
-From either of these tabs, click the *Open Session View* icon for an event or session.
-Labeled below are 1) the *Sessions* tab, and 2) the *Open Session View* icon:
+From either of these tabs, click the *Open Session View* button for an event or session.
+Labeled below are *1.* the *Sessions* tab, and *2.* the *Open Session View* button:
```
joepeeples, line 54: I don't think the number labels are really necessary here, or even the image itself: Elastic's [guidelines](https://docs.elastic.dev/content-design/tech-docs/writing-about-ui#callouts-on-screenshots) ("Callouts on screenshots") suggest labels when an image is "very complex," but this image is pretty simple. The Sessions tab is no more difficult to find than any other tab throughout the app, and the Open Session View icon was already identified in the previous image (session-view-action-icon-detail.png).

If we did want to keep the labels and image, they'd probably look better and more consistent if the labels were formatted like the Session View UI section below (basically a numbered list).

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,37 @@ Session View includes an additional badge not pictured above:
...
+[role="screenshot"]
+image::images/session-view-output-badge.png[The Output badge,width=80%,height=80%]
+
+[[session-view-output]]
+[discrete]
+=== Terminal output view UI
+
+beta::[]
+
+Terminal output view helps provide a more complete understanding of commands executed by users or adversaries, and assists with auditing and compliance objectives. It allows you to review a session's TTY output, in the context of the commands which generated it. It includes both TTY output, and shell commands (if the shell was in echo mode).
```
joepeeples, line 109: Editing out some filler words and comma splices. Don't know if "original commands" is the right phrase, but "commands which generated it" feels too wordy.

Also, i'm not sure about the last part "as typed -- provided the shell etc": The way it's phrased, it makes me wonder, well what if it _wasn't_ in echo mode? Wouldn't most bad actors turn off echo to hide their actions? Also "as typed" implies that every command is being manually typed by a human sitting at a keyboard, but I assume most malicious commands are scripted, API'd, or somehow automated. I might not understand the context of this feature, though -- maybe related to needing to explain what "TTY" means, and what exactly this output represents?
```suggestion
Terminal output view provides a more complete understanding of commands executed by users or adversaries, and assists with auditing and compliance. It allows you to review a session's TTY output in the context of the original commands. It includes both TTY output and shell commands (if the shell was in echo mode).
```

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,37 @@ Session View includes an additional badge not pictured above:
...
+and select the *Include session data* and *Capture terminal output* options.
+
+You can configure several additional settings by clicking *Advanced settings* at the bottom of the page:
+
+* Select the maximum number of kilobytes of output to record from a single process by modifying `linux.advanced.tty_io.max_kilobytes_per_process`. Default: 512 KB. For example, with the default value, a process' output will not be recorded after the 512th KB.
+* Select the maximum number of kilobytes of output to send to {es} as a single event by modifying `linux.advanced.tty_io.max_kilobytes_per_event`. Default: 512 KB. For example, with the default value, the maximum amount of output that will be sent to {es} as a single event is 512 KB. Additional data will be captured as a new event.
+* Select the maximum duration in seconds for which output is batched by modifying `linux.advanced.tty_io.max_event_interval_seconds`. Default: 30 seconds. For example, with the default value, any recorded output will be sent to {es} every 30 seconds (unless it first exceeds the `max_kilobytes_per_event` value, in which case it might be sent sooner).
```
joepeeples, line 120: For config settings like these, a table or definition list works better than full narrative sentences, which can get repetitive and unreadable. Also the examples seem a little unnecessary or kinda obvious, so maybe rephrase them more like statements.

Something like:

* `linux.advanced.tty_io.max_kilobytes_per_process` - The maximum amount of output (in kilobytes) to record from a single process. Process output will not be recorded after reaching the maximum. Defaults to 512 KB.

* etc.

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,37 @@ Session View includes an additional badge not pictured above:
...
+
+[role="screenshot"]
+image::images/session-view-output-viewer.png[Terminal output view,width=100%,height=100%]
+
+1. The search bar. Use to find and highlight search terms within the current session.
+The left and right arrows allow you to navigate through search results.
+2. The right-side scroll bar. Use along with the bottom scroll bar to navigate output data that doesn't fit on a single screen.
```
joepeeples, line 127: This item feels really unnecessary: it sounds like these scroll bars work the same as every other scroll bar in every other app, so I don't know why we're calling attention to them. If we're concerned that users might not immediately realize they can scroll through output data, it's probably more appropriate to mention that as part of describing the general UI.

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,37 @@ Session View includes an additional badge not pictured above:
...
+3. The playback controls: *Start*, *Prev*, *Play*, *Next*, *Last*, and the progress bar. Use to advance or rewind the session's commands and output. Click anywhere on the progress bar to jump to that part of the session. The marks on the bar represent processes that generated output. Click them or the *Prev* and *Next* buttons to skip between processes.
```
joepeeples, line 128: Probably don't need to name all the buttons just for the sake of naming them, especially since they appear to be common playback controls that most users are very likely familiar with from media players, etc.
```suggestion
3. Playback controls and progress bar. Use to advance or rewind the session's commands and output. Click anywhere on the progress bar to jump to that part of the session. The marks on the bar represent processes that generated output. Click them or the *Prev* and *Next* buttons to skip between processes.
```

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,37 @@ Session View includes an additional badge not pictured above:
...
+4. The *Fit screen*, *Zoom in*, and *Zoom out* buttons. Use to adjust the text size. Use Session view's *Fullscreen* button (located next to the *Close session viewer* button) to better fit output with long lines, such as for graphical programs like `vim`.
```
joepeeples, line 129: The last sentence about the **Fullscreen** button is kind of confusing because that button isn't even in the screenshot above. I think I get why it's here, but maybe reformatting as a Tip might work better, since then it'll be separate from the rest of the description of # 4 from the diagram.

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +93,37 @@ Session View includes an additional badge not pictured above:
...
+=== Terminal output view UI
+
+beta::[]
+
+Terminal output view helps provide a more complete understanding of commands executed by users or adversaries, and assists with auditing and compliance objectives. It allows you to review a session's TTY output, in the context of the commands which generated it. It includes both TTY output, and shell commands (if the shell was in echo mode).
+
+To enable Terminal output data capture:
+. Go to *Manage* -> *Policies*, then select one or more of your {elastic-defend} integration policies to edit.
+. On the *Policy settings* tab, scroll down to the Linux event collection section near the bottom of the page
+and select the *Include session data* and *Capture terminal output* options.
```
joepeeples, line 114: List isn't rendering correctly -- probably needs a blank line between the lead-in and the first list item.

joepeeples (commented):
Good content, added some comments, thank you!

One general thing: noticed that "terminal" is both lowercase and uppercase in various spots throughout. I couldn't tell if there was a specific nuance behind that, but I'd opt for lowercase overall.

docs/detections/session-view.asciidoc
```diff
@@ -16,20 +16,27 @@ and investigating session activity on your Linux infrastructure and understandin
 * *Process and event telemetry:* Process information included in the Linux logical event model.
 * *Nested sessions:* Sessions started by processes descended from the entry session.
 * *Alerts:* Alerts in the context of the processes which caused them.
+* *TTY output:* Terminal output associated with each process in the session.
```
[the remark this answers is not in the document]
benironside: I think this is out of scope. TTY stands for teletypewriter, which is kind of a whole long story about the development of computing. In Linux world most people just say TTY, and the longer version wouldn't add any information.

docs/detections/session-view.asciidoc
benironside: This is a great point, hilariously stated, and I am making a note of it, though I'm not going to change it right now.

docs/detections/session-view.asciidoc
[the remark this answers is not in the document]
benironside: There was some discussion with the AWP team about adding more background info about Linux terminals, but we decided it's out of scope for this release.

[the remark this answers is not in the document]
mitodrummer: If it helps we can just omit the use of the word TTY in the docs, and just refer to it as "Terminal output"

docs/detections/session-view.asciidoc
mitodrummer: There is a difference with the way the scroll area works in this context. It doesn't scroll the text content in the terminal, it is a scroll area to vertically/horizontally position the inner terminal window (which is statically sized to the native host session size, e.g 120x80 (character columns/rows). If a host session is large, e.g 200x120 or something, it will exceed the bounds of the container, and so the combination of this scroll area, the zoom in / out and the fullscreen controls allow you to size this area so that it's native session fits the scrollpane. You'll notice if you hit the "fit to screen" button (when not fullscreen) for a session that has a natively large size, it will shrink the text inorder for the window to fit in the smaller scroll area.

docs/detections/session-view.asciidoc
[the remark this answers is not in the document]
joepeeples: Thanks @mitodrummer and @benironside, i think "terminal output" is a lot clearer. I was mostly trying to figure out if TTY was used in a uniquely meaningful way here, or if it was more used as jargon.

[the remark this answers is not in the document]
mitodrummer: Yea TTY is more of an underlying concept, so it's probably best to call this Terminal output across the board, product / documentation wise. That being said we have ECS fields to describe the "tty" for an interactive session (interactive = a session which has a controlling tty. see: https://www.elastic.co/guide/en/ecs/8.5/ecs-process.html#field-process-tty and https://www.elastic.co/guide/en/ecs/8.5/ecs-process.html#interactive

docs/detections/session-view.asciidoc
joepeeples: Thanks for explaining, @mitodrummer! I definitely think it's worth mentioning how the user can scroll, resize, etc. in the UI, and some of these details would be helpful to mention. 

@benironside if there's a way to summarize that in the current description, I think that'd work and it would be clearer why we're saying "look, a scroll bar." If it's too much info to summarize easily, then it might justify a separate section on navigating terminal session sizes, etc.

docs/detections/session-view.asciidoc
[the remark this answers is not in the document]
mitodrummer: https://www.baeldung.com/linux/terminal-shell-tty-vs-console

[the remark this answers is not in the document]
mitodrummer: Popped up on my feed 👨‍💻

mitodrummer (approved):
LGTM!

m-sample:
pls pardon my tardiness - I will have completed a review on this today.

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +90,40 @@ Session View includes an additional badge not pictured above:
...
+
+* `linux.advanced.tty_io.max_kilobytes_per_process`: The maximum number of kilobytes of output to record from a single process. Default: 512 KB. Process output exceeding this value will not be recorded.
+* `linux.advanced.tty_io.max_kilobytes_per_event`: The maximum number of kilobytes of output to send to {es} as a single event. Default: 512 KB. Additional data is captured as a new event.
+* `linux.advanced.tty_io.max_event_interval_seconds`: The maximum interval (in seconds) during which output is batched. Default: 30 seconds. Output will be sent to {es} at this interval (unless it first exceeds the `max_kilobytes_per_event` value, in which case it might be sent sooner).
+
+[role="screenshot"]
+image::images/session-view-output-viewer.png[Terminal output view]
+
+1. Search bar. Use to find and highlight search terms within the current session.
+The left and right arrows allow you to navigate through search results.
```
m-sample, line 124: Detail on why search may behave unexpectedly would help as well as emphasis that this is only search over terminal output held in the browser (ie it is not an Elasticsearch query - @mitodrummer pls correct me if I'm wrong).  E.g.:

"Use to find and highlight search terms within the current session's terminal output held by the browser.   Note that because terminal output may include terminal codes for colorizing the text and moving the cursor, some searches may not find a match.  For example, if one is searching output for a command like "sudo" (echoed command input) and the user mis-entered "sudo" as "sdo", and moved the cursor back to insert the missing "u", the terminal output search for "sudo" will not match.  This challenge can also be present in visual editors like vi and emacs where cursor movement for editing is commonplace"

m-sample (changes requested):

docs/detections/session-view.asciidoc
```diff
@@ -84,4 +90,40 @@ Session View includes an additional badge not pictured above:
...
+. On the *Policy settings* tab, scroll down to the Linux event collection section near the bottom of the page
+and select the *Include session data* and *Capture terminal output* options.
+
+You can configure several additional settings by clicking *Advanced settings* at the bottom of the page:
+
+* `linux.advanced.tty_io.max_kilobytes_per_process`: The maximum number of kilobytes of output to record from a single process. Default: 512 KB. Process output exceeding this value will not be recorded.
+* `linux.advanced.tty_io.max_kilobytes_per_event`: The maximum number of kilobytes of output to send to {es} as a single event. Default: 512 KB. Additional data is captured as a new event.
+* `linux.advanced.tty_io.max_event_interval_seconds`: The maximum interval (in seconds) during which output is batched. Default: 30 seconds. Output will be sent to {es} at this interval (unless it first exceeds the `max_kilobytes_per_event` value, in which case it might be sent sooner).
+
+[role="screenshot"]
```
m-sample, line 120: Do we have any docs on writing alerting rules based on matching on terminal output?    

It is important to note that when writing Alerting Rules that query Elasticsearch for matching terminal output that multi-line matching will not behave consistently because the output text may be stored in more than one output event.  For example,  Foo\nBar - that is, `Foo` on one line followed by `Bar` on the next line - will not behave consistently because Bar might be in output event after Foo.   Elastic tries to ensure a given line of terminal output is contained in exactly one output event to ensure in-line matches work as expected (beware of terminal color codes).   However, very long lines may be split across more than one output event.

benironside: We do not have any docs specifically about alerting rules based on terminal output, yet. But we could mention something here about the issue you outlined. 

Any thoughts on how to work around that limitation? I guess one thing would be to increase the maximum size of output events to reduce the number of unexpected newlines, but that would still be inconsistent.

m-sample: Yes, increasing the max event buffering size and 30 second capture to be larger (60s?) create bigger events that are more likely to work for multi-line matches but there are still chances for inconsistency.   Also, IIRC we flush the output immediately if the program writing to the terminal changes - e.g. I ctrl-Z a program that is producing output and the shell then echos a prompt (the shell is another program writing to the terminal, causing a flush), then put the original program back into the foreground with `fg`.   THis will break up the output of the original program over at least 2 events.

If this important for some customers, the best way is to post process these output events to group all the output events of a given process, and while they're at it ideally scrub out terminal control codes for color, etc.

karanbirsingh-qasource:
Hi @jmikell821 

we have validated the above shared preview links are documentation are good to go.

**Screen-Shot**

image

docs/detections/session-view.asciidoc
mitodrummer: You are correct @m-sample, though, for the browser side "plain text" search, I strip out all ansi chars on a a line before matching the user's search against it. This is required so that I get an accurate string index position for each match, so that when I seek to this "line" in the tty player, it's able to highlight the correct match (if there are multiple on a single line). That being said, you are correct in that if a user mistypes a command and then hits backspace to correct it, the resulting output text will not be the "corrected" version.

e.g
User types,    sdo
They hit backspace twice, then type udo to fix their typo.
The resulting output will look something simlilar to:
sdo\b\budo

The \b is an escape code for backspace, so you can see why searching for the corrected command will not match.

That being said, most of the formating ansi codes for color and text weight would be stripped out and should not interfere with most plain text searches. This is in contrast to using KQL or EQL to match on values in process.io.text as it will be the un altered version of the value that will potentially contain many escape sequences between words in the output text. Here is a link to a resource which describes these codes. https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797

docs/detections/session-view.asciidoc
mitodrummer: As for rules that match against process.io.text, it is advised that all rules match with leading/trailing wildcards (*).

e.g
```process.io.text: *hello world*```

EQL equivalent:
```process where process.io.text : "*hello world*"```

This will ensure a less strict rule which will have a much better chance of firing. Without the * wildcards, it would try to match an event with exactly the string "hello world" which in most cases will not match on anything, since an output event can have multiple lines in it. This also opens up another potential bit of confusion for the end user. E.g if their event size is quite large. e.g 512kb (due to some long cat'd file), there will be a single alert for this document no matter how little the text we are matching against. This could make it difficult to figure out which part of the proces.io.text value actually triggered the alert. We may add some UX down the road to help with this. e.g showing alerts where event.action: 'text_output' inside the tty player itself (potentially highlighting the bit of text responsible for the alert). Reducing the timeout and batch size configs could help limit the amount of text per event, thus alleviating this issue.

File: docs/detections/session-view.asciidoc [hunk 25]
```diff
@@ -84,4 +90,53 @@ Session View includes an additional badge not pictured above:
...
+
+[discrete]
+[[terminal-output-limitations]]
+==== Terminal output limitations for search and alerting
+You should understand several current limitations before building rules based on terminal output data:
+
+* Terminal output that appears in the `process.io.text` field includes https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797[ANSI codes] that represent, among other things, text color, text weight, and escape sequences. This can prevent EKS queries from matching as expected. Queries of this data will have more success matching single words than more complex strings.
+* Queries of this data should include leading and trailing wildcards (for example `process where process.io.text : "*sudo*"`), since output events typically include multiple lines of output.
+* The search functionality built in to terminal output view is subject to similar limitations. For example, if a user accidentally entered `sdo` instead of `sudo`, then pressed backspace twice to fix the typo, the recorded output would be `sdo\b\budo`. This would appear in the terminal output view as `sudo`, but searching terminal output view for `sudo` would not result in a match.
```
joepeeples, line 140:
```suggestion
* The search functionality built into terminal output view is subject to similar limitations. For example, if a user accidentally entered `sdo` instead of `sudo`, then pressed backspace twice to fix the typo, the recorded output would be `sdo\b\budo`. This would appear in the terminal output view as `sudo`, but searching terminal output view for `sudo` would not result in a match.
```

opauloh (commented):
LGTM

PR #2567, opened 2022-10-12
[DOCS] Bulk edit rules updates (rule schedules, actions)

### LINKED ISSUE #2441: [DOCS] Bulk edit rule actions
## Description 

Per https://github.com/elastic/kibana/pull/138900, users can bulk edit actions on multiple rules.

## Images

NOTE: UI copy is still being edited; these are preliminary images.

New item in "Bulk actions" menu:
image

New flyout:
image

## Notes

- Selecting an Actions frequency applies that to all actions (both new and existing) on all selected rules. That means you can basically overwrite the frequency on existing actions, even without checking the "Overwrite all selected rules actions" box. (That checkbox appears to apply only to the actions themselves, not frequency.) 
- API docs probably also need updating: https://www.elastic.co/guide/en/security/master/bulk-actions-rules-api.html#bulk-edit-object-schema

### LINKED ISSUE #2453: [DOCS] Bulk edit rule schedules
## Description 

- With work done for https://github.com/elastic/security-team/issues/2172 in PR https://github.com/elastic/kibana/pull/140166, users can now bulk edit schedule of rules (interval -runs every- and lookback time)

## Images 

**New menu item for updating schedules in bulk**

image

**Modal warning that prebuilt rules are not editable**

image

**Flyout for updating schedules in bulk**

image

## Notes

- The bulk schedule form provides validation for the `interval` and `lookback` fields. The values must be a positive integer greater than `0`. If the user tries to force a negative number or 0, the input will be automatically corrected to the minimum value of `1`,

joepeeples:
Related issues — Streamlined into a single PR:
* Resolves #2441
* Resolves #2453 

Preview: [Manage detection rules | Modify existing rules settings](https://security-docs_2567.docs-preview.app.elstc.co/guide/en/security/master/rules-ui-management.html#edit-rules-settings)


nastasha-solomon (approved):

docs/detections/rules-ui-manage.asciidoc
```diff
@@ -120,6 +120,11 @@ If you try to bulk edit with both prebuilt and custom rules selected, the action
 * Bulk edit multiple rules: Select the rules you want to edit, then select an action from the *Bulk actions* menu:
 ** *Index patterns*: Add or delete the index patterns used by all selected rules.
 ** *Tags*: Add or delete tags on all selected rules.
+** *Add rule actions*: Add <<rule-notifications,notification actions>> on all selected rules.
++
+NOTE: The action frequency you select is applied to all actions (both new and existing) on all selected rules. If you don't want to change the frequency of existing actions, update the rules separately.
```
benironside, line 125:
```suggestion
NOTE: The action frequency you select applies to all actions (both new and existing) on all selected rules. If you don't want to change the frequency of existing actions, update the rules separately.
```

benironside (commented):
One trivial suggestion for your consideration, otherwise LGTM

docs/detections/rules-ui-manage.asciidoc
```diff
@@ -120,6 +120,11 @@ If you try to bulk edit with both prebuilt and custom rules selected, the action
 * Bulk edit multiple rules: Select the rules you want to edit, then select an action from the *Bulk actions* menu:
 ** *Index patterns*: Add or delete the index patterns used by all selected rules.
 ** *Tags*: Add or delete tags on all selected rules.
+** *Add rule actions*: Add <<rule-notifications,notification actions>> on all selected rules.
```
vitaliidm, line 123: Should we note that `Add rule actions` can be applied to prebuilt rules as well?

Looking at this note above the list of bulk edit, it might be a bit confusing, in which cases action will affect prebuilt rules, in which not.

Screenshot 2022-10-13 at 16 03 44

> If you try to bulk edit with both prebuilt and custom rules selected, the action will affect only custom rules.

Also, it looks like this note doesn't reflect[ appearance of modal window,](https://github.com/elastic/kibana/issues/124918#issuecomment-1199146225), which happens when user selects both custom and prebuilt rules. And can decide which action to take: edit only custom rules OR cancel action at all

joepeeples: Thanks @vitaliidm! I agree, that note isn't quite accurate, so I revised it:

> If you try to bulk edit with both prebuilt and custom rules selected, the action will affect only the rules that can be modified.

The modal's UI text already does a good job of explaining the situation and what options the user has, so I don't think it's necessary to comment any further in docs; there isn't much more we could say, so we'd probably just be repeating what the UI said.

vitaliidm (approved):
LGTM

karanverma-qasource:
Hi Team,

We have validated above PR and the changes are good to go. Hence, marking it as 'QA Validated.

Thanks !

PR #2620, opened 2022-10-24
[DOCS] Troubleshooting: no suggested values when creating exceptions

### LINKED ISSUE #2284: [DOCS] Troubleshooting: no suggested values when creating exceptions
## Description 

When creating detection rule exceptions, autocomplete might not provide Value suggestions. 

image

This is due the Kibana advanced setting `autocomplete:useTimeRange`, which limits the time range for looking up autocomplete values. With this setting turned on, in order to get suggested, values must exist within the time range set on the page's time picker.

image

Disabling this advanced setting could cause performance issues, so we only want to mention this as a troubleshooting issue: if users have trouble with creating exceptions, here's a possible reason. But if users aren't having any trouble with exceptions, they don't need to disable the advanced setting.

### Notes

- Include link to Kibana docs for advanced settings
- Include admonition about performance issues, etc.
- Backport to 7.17
- OK to prioritize for 8.5, nice to have for 8.4
- Related: https://github.com/elastic/kibana/issues/137709
- Related: https://github.com/elastic/kibana/issues/100584

joepeeples:
Resolves #2284.

Preview: [Troubleshooting | Detection rules | Rule exceptions](https://security-docs_2620.docs-preview.app.elstc.co/guide/en/security/master/ts-detection-rules.html#rule-exceptions-ts)


dhurley14 (approved):

docs/troubleshooting/detections/detection-rules.asciidoc
```diff
@@ -65,3 +65,19 @@ If you receive the following rule failure: `Bulk Indexing of signals failed: ind
...
+[discrete]
+[[rule-exceptions-ts]]
+=== Rule exceptions
+
+[discrete]
+[[rule-exceptions-autocomplete-ts]]
+.No autocomplete suggestions
+[%collapsible]
+====
+When creating detection rule exceptions, autocomplete might not provide suggestions in the *Value* field if the data doesn't exist in the current page's time range.
```
benironside, line 78:
```suggestion
When you create detection rule exceptions, autocomplete might not provide suggestions in the *Value* field if data doesn't exist in the current page's time range.
```
or maybe "the affected data"? As written I'm not sure what the "the" is referring to

benironside: or maybe "When you are creating"? As written it seems to say autocomplete is creating detection rules, which might be fine since I'm pretty sure readers will figure out what you actually mean pretty much immediately 🤷

benironside (commented):
LGTM, minor suggestions

docs/troubleshooting/detections/detection-rules.asciidoc
```diff
@@ -65,3 +65,19 @@ If you receive the following rule failure: `Bulk Indexing of signals failed: ind
...
+
+You can resolve this by expanding the time range, or by configuring {kib}'s autocomplete feature to get suggestions from your full data set instead of the time range. Go to *{kib}* -> *Stack Management* -> *Advanced Settings*, then turn off `autocomplete:useTimeRange`.
```
nastasha-solomon, line 80:
```suggestion
You can resolve this by expanding the time range, or by configuring {kib}'s autocomplete feature to get suggestions from your full data set instead. Go to *{kib}* -> *Stack Management* -> *Advanced Settings*, then turn off `autocomplete:useTimeRange`.
```

nastasha-solomon (approved):
Just one minor edit - otherwise, LGTM!

docs/troubleshooting/detections/detection-rules.asciidoc
joepeeples: Adding "you" to the first part sounds good, natural and conversational. 😄 Throw in a contraction for even more casual tone.

For "the" data, it's the specific data values used to provide autocomplete suggestions, not data in general. Maybe "the values" is more specific, and points back to the **Value** field:

```suggestion
When you're creating detection rule exceptions, autocomplete might not provide suggestions in the *Value* field if the values don't exist in the current page's time range.
```

jmikell821 (approved):
Love that we're adding to this section; LGTM, thanks!

PR #2732, opened 2022-11-18
[DOCS] New Terms rule type - allow multiple fields

### LINKED ISSUE #2718: [DOCS] New Terms rule type - allow multiple fields
## Description

The "new terms" rule type now supports up to 3 fields to check for new terms (previously only 1 field was available).

Related:
- https://github.com/elastic/kibana/issues/142862
- https://github.com/elastic/kibana/pull/143943
- https://github.com/elastic/kibana/pull/145707 — see videos in PR description

## Notes

- This affects several areas of the app (maybe more than this, just an initial list):
   - Create rules page - Selecting the fields to use
   - Alert details - Alert will display the new fields found
   - Maybe also Alert table? - New term fields and values will be part of the alert's data; confirm if any screenshots need updating, etc.
   - API docs
- Update: the only docs affected are the Create rules page; API docs will be updated in a separate issue

joepeeples:
Resolves #2718.

Preview: [Create a detection rule | Create a new terms rule](https://security-docs_2732.docs-preview.app.elstc.co/guide/en/security/master/rules-ui-create.html#create-new-terms-rule)


benironside (approved):

jmikell821 (approved):
LGTM!

docs/detections/rules-ui-create.asciidoc
```diff
@@ -286,7 +286,7 @@ alerts.
 +
 NOTE: You can use {kib} saved queries (image:images/saved-query-menu.png[Saved query menu,18,18]) and queries from saved Timelines (*Import query from saved Timeline*) as rule conditions.
 +
-.. Use the *Fields* menu to select a field to check for new terms.
+.. Use the *Fields* menu to select a field to check for new terms. You can also select up to three fields to detect a combination of new terms (for example, a `host.ip` and `host.id` that have never been observed together before).
```
vitaliidm, line 289: There is a limitation currently, when single document has more than 100 unique combination for new terms fields.
https://github.com/elastic/kibana/blob/main/x-pack/plugins/security_solution/server/lib/detection_engine/rule_types/new_terms/README.md#limitations-and-future-enhancements
Would be worth to mention it

joepeeples: @vitaliidm Thanks for this info, definitely will include it! Do you know what exactly happens if a document falls within this limitation? For example, will the rule still run, or will it fail? Will it still analyze some of the data (such as the first 100 combinations found)? Is there an error or a log message to indicate that some data wasn't analyzed?

vitaliidm: >  Do you know what exactly happens if a document falls within this limitation? For example, will the rule still run, or will it fail? Will it still analyze some of the data (such as the first 100 combinations found)? Is there an error or a log message to indicate that some data wasn't analyzed?

rule will still run, missing results might happen if number of unique combination too large.
there are no indications whether it happened or not.
In one of previous versions of implementation rule was failing, when number of combination is greater than 100. But we decided to keep current implementation, as it would allow rule to finish run.

At this point there are no workarounds or alternatives to it.
We have slower in performance implementation, that doesn't have this weakness. But has some other [limitations](https://github.com/elastic/kibana/issues/142862#issuecomment-1316865039)

joepeeples: @vitaliidm I added the note below to mention the limitation. Please let me know if it needs any adjustment, thanks!

> IMPORTANT: When checking multiple fields, results are limited to no more than 100 unique combinations; any additional combinations will not be detected. If you expect your data to produce more than 100 unique new terms combinations, consider checking other fields to narrow down the search.

vitaliidm: > IMPORTANT: When checking multiple fields, full results can be produced only for documents with no more than 100 unique combinations of values in these fields. If you expect your data to produce more than 100 unique new terms combinations per document, consider checking other fields to narrow down the search.

vitaliidm (approved):

nastasha-solomon (approved):

karanverma-qasource:
Hi @joepeeples  ,

We tested the linked docs and found that the doc is correctly updated as per the latest UI. So we are good to go ahead and merge the changes.

Marking this ticket as 'QA validated'.

Thanks!

PR #2751, opened 2022-11-29
Add rule docs for 8.6 rule changes

terrancedejesus:
Kibana Security Doc updates for rule changes in 8.6.


.gitignore
```diff
@@ -1,3 +1,6 @@
 .DS_Store
 docs/html_docs
 /html_docs
+
+# development files
+*launch.json*
```
terrancedejesus, line 6: Added this to ignore debugging file for Python.

docs/detections/prebuilt-rules/prebuilt-rules-changelog.asciidoc
```diff
@@ -5,6 +5,9 @@ The following lists prebuilt rule updates per release. Only rules with
 significant modifications to their query or scope are listed. For detailed
 information about a rule's changes, see the rule's description page.
 
+[float]
```
terrancedejesus, line 8: This should list all the files that have been changed, but does not. I will need to look into this.

prebuilt-rules-scripts/generate.py
```diff
@@ -15,7 +15,7 @@
 PREBUILT_RULES = ROOT.joinpath('prebuilt-rules-scripts')
 GENERATED_ASCII = ROOT.joinpath('generated-ascii-files')
 DEFAULT_KIBANA_RULES_DIR = str(Path().joinpath('x-pack', 'plugins', 'security_solution', 'server', 'lib',
-                                               'detection_engine', 'rules', 'prepackaged_rules'))
+                                               'detection_engine', 'prebuilt_rules', 'content', 'prepackaged_rules'))
```
terrancedejesus, line 18: Updated this so it points to the Kibana FS rules after recent change.

Reference -  https://github.com/elastic/kibana/pull/142950

terrancedejesus:
These security docs do not seem to be correct as a result of a failed backport in Kibana. Since our process requires us to checkout Kibana branch 8.6 and then create docs from there, changes not backported would not be reflected in doc updates.

Reference - https://github.com/elastic/kibana/pull/146402#issuecomment-1330748190

PR #2780, opened 2022-12-12
Duplicate rules API update

nkhristinin:
Preview [here](https://security-docs_2780.docs-preview.app.elstc.co/guide/en/security/master/bulk-actions-rules-api.html#bulk-duplicate-object-schema)


joepeeples (approved):
LGTM!

Edit: Actually noticed one line that could probably be removed, but that's all. Thanks!

PR #2782, opened 2022-12-12
[DOCS] Move detections and alerts terminology to glossary

lcawl:
This PR removes the terminology from https://www.elastic.co/guide/en/security/master/detection-engine-overview.html#det-engine-terminology since it is moved to the Elastic Glossary in https://github.com/elastic/stack-docs/pull/2288. The goal is to have a single glossary that contains terms for all Elastic products.


docs/getting-started/siem-ui.asciidoc
```diff
@@ -161,7 +161,7 @@ There are also tabs for viewing and investigating specific types of data:
 {apm-app-ref}/apm-getting-started.html[Elastic APM] are monitored by default)
 * TLS: handshake details
 * Anomalies: anomalies discovered by <<machine-learning, machine learning jobs>>
-* <<det-engine-terminology, External alerts>>: alerts received from
+* External alerts: alerts received from
 external monitoring tools, such as Elastic Endpoint Security
```
jmikell821, line 165: @joepeeples I'm checking the terminology of `Elastic Endpoint Security` here. AFAIK, we're not using this term anymore except for when referring to the endpoint rule -- but double checking my sanity.

jmikell821 (commented):
Thanks for pulling this together @lcawl! Just a few comments on my end and a question for @joepeeples. Can we see the Elastic Glossary draft when it's pulled together please?

docs/getting-started/siem-ui.asciidoc
joepeeples: @jmikell821 Correct, "Elastic Endpoint Security" is now "Elastic Defend." It's a bit moot in this case, though, because this file `siem-ui.asciidoc` is in the old (retired?) [SIEM Guide](https://www.elastic.co/guide/en/siem/guide/current/siem-ui-overview.html) book; I don't think this content will ever show up in current Elastic Security docs. That's why it wasn't revised when the integration was renamed, or re-renamed.

For the sake of this PR, no further changes needed; we can decide separately if we need to follow up on the terminology.

jmikell821: @joepeeples true. Customers won't see it unless they're looking at `7.8` and earlier -- so I'm not too worried about it. Thanks!

lcawl: The "Elastic Endpoint Security" terminology was also used in the "endpoint exception" term that I carried over into the glossary: https://www.elastic.co/guide/en/elastic-stack-glossary/current/terms.html#e-glos. If that's not a useful term anymore, let me know and we can correct or remove that entry.

lcawl:
> Can we see the Elastic Glossary draft when it's pulled together please?

@jmikell821 The detections-related terms are now visible publicly in https://www.elastic.co/guide/en/elastic-stack-glossary/current/terms.html

docs/getting-started/siem-ui.asciidoc
joepeeples: @lcawl Actually that particular instance is OK as is, too -- another case of "It Depends"! 🙃 The "Elastic Endpoint Security rule" mentioned in that glossary entry has not yet been renamed to match the integration's name changes, so it's another rare case where we don't need to change it in docs (yet?). If they ever do rename the detection rule, we can update the glossary at that time.

jmikell821 (approved):

PR #2946, opened 2023-01-27
fix air-gapped artifact download commands

stevewritescode:
During testing I found an error in the sample command we provide as part of these new 8.6 docs. The previous script downloaded some files to an incorrect file path which caused Elastic Endpoint to ignore the artifact update.

Also during testing, I found an Endpoint bug https://github.com/elastic/endpoint-dev/issues/12382 which causes Endpoint to reject some requests from nginx artifact servers. So I updated the example configuration with a workaround.

I tested these updated commands as well as the nginx configuration and httpd configuration on a Windows 10 VM and confirmed Endpoint downloads and installed the updated artifacts using this download script 🚀

### DIFF
--- a/docs/getting-started/offline-endpoint.asciidoc
+++ b/docs/getting-started/offline-endpoint.asciidoc
@@ -106,9 +106,15 @@ This example script starts an Nginx Docker image and configures it as a file ser
 [source,sh]
 ----
 cat > nginx.conf << EOF
+# set compatible etag format
+map $sent_http_etag $elastic_etag {
+  "~(.*)-(.*)" "$1$2";
+}
 server {
   root /app/static;
-  location / {}
+  location / {
+    add_header ETag "$elastic_etag";
+  }
 }
 EOF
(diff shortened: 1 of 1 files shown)


stevewritescode:
Updated the description.

joepeeples (approved):
LGTM, thanks!

@stevewritescode Do you think we'll need to update these docs again once the bug in https://github.com/elastic/endpoint-dev/issues/12382 is resolved, and the workaround is no longer needed? Should be a simple docs update if needed, just wondering if it's something we should keep an eye on.

stevewritescode:
@joepeeples - yes, hopefully we can simplify the docs once the workaround is no longer needed, at least for future releases. Once that bug is merged I'll plan to circle back again with a docs PR for it.

stevewritescode:
I'm doing one final validation this morning on my test stack deployment and will merge after that.

stevewritescode:
Confirmed that Endpoint 8.6.0 can download, validate, and apply global artifacts from a local nginx file server using the documented commands. Merging.

Issue #2986, opened 2023-02-15
[Cloud Posture] Add note regarding Kibana default space


ari-aviran:
## Description 

Cloud security posture only works when installed in the default Kibana space. We need to add a note to the docs that emphasizes this fact.

## Acceptance Test Criteria

As a user browsing the cloud security posture docs I need to understand that I can only install it on the default Kibana space.

PR #2987, opened 2023-02-15
[Cloud Posture] Add note regarding supported Kibana space

ari-aviran:
Add a note to the "Get started with KSPM" page regarding support for Kibana spaces - only the `Default` space is supported.

Resolves https://github.com/elastic/security-docs/issues/2986


ari-aviran:
For some reason I can't add reviewers nor labels.
@tinnytintin10 @benironside could you please review this change?

In addition this needs to be back-ported to 8.7, 8.6 and 8.5, could you please add the appropriate labels?

Thanks

Issue #2936, opened 2023-01-26
[BUG] Exception "matches" operator example is incorrect and misleading


peluja1012:
**Description**:
In our documentation for wildcard/matches exception operator, we include an example that wouldn't yield expected results. More specifically, we say they can create a `matches` exception with a value of `C:\path\*\app.exe`. This won't match the results the user expects because the backslash (`\`) character is a special character and they would need to escape it. We have two requests:

1. Could we please update the example value to `C:\\path\\*\\app.exe`?
2. Could we pleased add a note saying that the backslash `\` character needs to be escaped (`\\`)?

cc @marshallmain 

**Doc URL link or topic name**:
https://www.elastic.co/guide/en/security/current/add-exceptions.html#detection-rule-exceptions

<img width="612" alt="image" src="https://user-images.githubusercontent.com/616158/214922898-ebceabb1-07ee-45a0-862f-7c268ecd87d5.png">

nastasha-solomon:
Thanks for filing this bug, @peluja1012 ! Are there any other special characters that need to be escaped and that users need to be aware of?

PR #3018, opened 2023-02-22
[BUG] Exception "matches" operator example is incorrect and misleading

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/2936.

Preview [here](https://security-docs_3018.docs-preview.app.elstc.co/guide/en/security/master/add-exceptions.html#detection-rule-exceptions) - see step 2b


docs/detections/add-exceptions.asciidoc
```diff
@@ -69,7 +69,9 @@ Add conditions that define when the exception prevents alerts:
...
 =======
-    * `matches` | `does not match` — Allows you to use wildcards in *Value*, such as `C:\path\*\app.exe`. Available wildcards are `?` (match one character) and `*` (match zero or more characters). The selected *Field* data type must be {ref}/keyword.html#keyword-field-type[keyword], {ref}/text.html#text-field-type[text], or {ref}/keyword.html#wildcard-field-type[wildcard].
+    * `matches` | `does not match` — Allows you to use wildcards in *Value*, such as `C:\\path\\*\\app.exe`. Available wildcards are `?` (match one character) and `*` (match zero or more characters). The selected *Field* data type must be {ref}/keyword.html#keyword-field-type[keyword], {ref}/text.html#text-field-type[text], or {ref}/keyword.html#wildcard-field-type[wildcard].
++
+NOTE: Some characters must be escaped with a backslash, such as `\"` for a quotation mark and `\\` for a literal backslash. Windows paths must be divided with double backslashes (for example, `C:\\Windows\\explorer.exe`), and paths that already include double backslashes might require four backslashes for each divider.
```
peluja1012, line 74: We don't believe quotes need to be escaped. Those will be escaped automatically in the backend. The characters that would need to be escaped explicitly that we know of are `\`, `*`, and `?`. If `*` is not escaped it would interpreted as `matches zero or more characters`. If `?` is not escaped it would be interpreted as `matched one character`.

nastasha-solomon: Oops, forgot to remove the first part about escaping the quotes. Let me know the following revision is accurate or needs additional changes:

```suggestion
NOTE: Some characters must be escaped with a backslash, such as `\\` for a literal backslash, `\* for an asterisk, and \? for a question mark. If `*` is not escaped, it is interpreted as `matches zero or more characters`. Similarly, if `?` is not escaped it is interpreted as "matched one character". Also note that Windows paths must be divided with double backslashes (for example, `C:\\Windows\\explorer.exe`), and paths that already include double backslashes might require four backslashes for each divider.      
```

peluja1012 (approved):

docs/detections/add-exceptions.asciidoc
nastasha-solomon: I agree, no need to duplicate the info. One more question: should the star char (`*`) be called the asterisk character or the wildcard character? I'm not sure which term is more accurate to use in the context of this sentence.

peluja1012: I think that technically both `*` and `?` are considered "wildcard operators".

peluja1012: In the context of this sentence, I think asterisk is correct because they would want to use `\*` if they want the value to be interpreted as an asterisk.

docs/detections/add-exceptions.asciidoc
```diff
@@ -69,7 +69,9 @@ Add conditions that define when the exception prevents alerts:
...
 =======
-    * `matches` | `does not match` — Allows you to use wildcards in *Value*, such as `C:\path\*\app.exe`. Available wildcards are `?` (match one character) and `*` (match zero or more characters). The selected *Field* data type must be {ref}/keyword.html#keyword-field-type[keyword], {ref}/text.html#text-field-type[text], or {ref}/keyword.html#wildcard-field-type[wildcard].
+    * `matches` | `does not match` — Allows you to use wildcards in *Value*, such as `C:\\path\\*\\app.exe`. Available wildcards are `?` (match one character) and `*` (match zero or more characters). The selected *Field* data type must be {ref}/keyword.html#keyword-field-type[keyword], {ref}/text.html#text-field-type[text], or {ref}/keyword.html#wildcard-field-type[wildcard].
++
+NOTE: Some characters must be escaped with a backslash, such as `\\` for a literal backslash, `\*` for an asterisk, and `\?` for a question mark. . Windows paths must be divided with double backslashes (for example, `C:\\Windows\\explorer.exe`), and paths that already include double backslashes might require four backslashes for each divider.
```
joepeeples, line 74:
```suggestion
NOTE: Some characters must be escaped with a backslash, such as `\\` for a literal backslash, `\*` for an asterisk, and `\?` for a question mark. Windows paths must be divided with double backslashes (for example, `C:\\Windows\\explorer.exe`), and paths that already include double backslashes might require four backslashes for each divider.      
```

joepeeples (commented):
One smol typo fix, otherwise LGTM!

docs/detections/add-exceptions.asciidoc
nastasha-solomon: Good catch, thanks!

jmikell821 (approved):
LGTM 👍

Issue #2981, opened 2023-02-14
Osquery query parameters can act as placeholders that will be replaced with event data


nastasha-solomon:
## Description 

Users can construct dynamic Osquery queries by incorporating placeholder fields into their queries. Placeholder fields use mustache syntax (`{{}}`) and are injected with real alert/event values when the query is executed. Dynamic queries are supported for: 
- **Rule investigation guides using Osquery queries** - In order for placeholder fields to be replaced by alert values, the investigation guide must be run from an alert generated by the rule.
- **Osquery Response Actions on rules** - Placeholder fields will only work if the rule has generated alerts. 
- **Live queries or query packs that are run on alerts or events** - Placeholder fields will only work if the data they're seeking exists in the alert or event's document. If that data does not exist, users will get the following error message (confirm this message):
<img width="347" alt="Screenshot 2023-03-05 at 3 57 36 PM" src="https://user-images.githubusercontent.com/79124755/222986123-a735fa5c-7ff9-4b54-b935-86b71e575cb2.png">

Example:
In this query, the user is saying: select all columns from the `os_version` table where name is the value in the host.name field. 

```
SELECT * FROM os_version WHERE name = `{{host.name}}`
```

Related:
- https://github.com/elastic/security-team/issues/5348
- https://github.com/elastic/kibana/pull/146598
- https://github.com/elastic/security-team/issues/5627: If a rule's custom investigation guide contains Osquery queries, user can add these queries to the rule's Osquery Response actions. (TBD) 

### Doc updates
- [x] Will need to create a new topic for this feature and add it to the Osquery TOC. The new topic should describe the following:
  - [x] Syntax that placeholder fields use (include an example) - refer to [these docs](https://www.elastic.co/guide/en/security/master/rules-ui-create.html#rule-action-variables) for example of how mustache syntax is used for alert notification fields.
  - [ ] What causes the `Skipped` status and what it means (see notes below for more info)
- Add notes in the following areas that placeholder fields can be added to queries. Link to the new topic for more information:
  - [x] Step 2 of [Add Osquery Response Actions to rules](https://www.elastic.co/guide/en/security/master/osquery-response-action.html#add-osquery-response-action)
  - [x] Step 3 of [Add live queries to an investigation guide](https://www.elastic.co/guide/en/security/master/invest-guide-run-osquery.html)
  - [x] Step 4 of [Run Osquery from alerts](https://www.elastic.co/guide/en/security/master/alerts-run-osquery.html)

## Notes
[...]

PR #3045, opened 2023-03-08
Placeholder fields and Osquery UI updates

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/2981.

Previews:
- [Use placeholder fields in Osquery queries](https://security-docs_3045.docs-preview.app.elstc.co/guide/en/security/master/osquery-placeholder-fields.html) - New topic
- [Add Osquery Response Actions to rules](https://security-docs_3045.docs-preview.app.elstc.co/guide/en/security/master/osquery-response-action.html#add-osquery-response-action) - Added note under step 2. 
- [Add live queries to an investigation guide](https://security-docs_3045.docs-preview.app.elstc.co/guide/en/security/master/invest-guide-run-osquery.html#add-live-queries-ig) - Added same note under step 3. 
- [Run Osquery from alerts](https://security-docs_3045.docs-preview.app.elstc.co/guide/en/security/master/alerts-run-osquery.html) - Added same note under step 4.


docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,20 @@
+[[osquery-placeholder-fields]]
+== Use placeholder fields in Osquery queries 
+
+Placeholder fields dynamically inject alert and event data into queries that you run. To create a placeholder field, you must wrap the field in double curly brackets (`{{}}`); you can use any field in an event or alert document.
```
nastasha-solomon, line 4:
```suggestion
Placeholder fields dynamically inject alert and event data into Osquery queries at runtime. Placeholder fields use mustache syntax and must be wrapped in double curly brackets (`{{}}`); you can turn any field in an event or alert document into a placeholder field. 
```

szwarckonrad:
CC @tomsonpl @patrykkopycinski

docs/osquery/osquery-index.asciidoc
```diff
@@ -6,4 +6,6 @@ include::invest-guide-run-osquery.asciidoc[][leveloffset=+1]
 
 include::alerts-run-osquery.asciidoc[][leveloffset=+1]
 
-include::view-osquery-results.asciidoc[][leveloffset=+1]
\ No newline at end of file
+include::view-osquery-results.asciidoc[][leveloffset=+1]
```
szwarckonrad, line 9: Comment test.

tomsonpl (approved):
LGTM 👍

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+* <<osquery-response-action,Osquery Response Action>>
++
+NOTE: Placeholder fields within an Osquery Response Action query will only work if the rule has generated alerts.
+* <<invest-guide-run-osquery,Investigation guides using Osquery queries>> 
+
+[float]
+[[placeholder-field-syntax]]
+=== Placeholder field syntax and requirements
+
+Placeholder fields use http://mustache.github.io/[mustache syntax] and must be wrapped in double curly brackets (`{{}}`). You can use any field within an event or alert document as a placeholder field.
```
joepeeples, line 18: All those brackets are a little hard to parse visually. How about something like:

```suggestion
Placeholder fields use http://mustache.github.io/[mustache syntax] and must be wrapped in double curly brackets (`{{example.field}}`). You can use any field within an event or alert document as a placeholder field. 
```

joepeeples (commented):
Good stuff! Just added a couple small suggestions, thanks!

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,30 @@
+[[osquery-placeholder-fields]]
+== Use placeholder fields in Osquery queries 
+
+Instead of hard-coding alert and event values into Osquery queries, you can use placeholder fields to dynamically pass this data into queries. Placeholder fields function like replaceable parameters. You can use placeholder fields to build flexible and re-usable queries.
```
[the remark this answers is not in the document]
benironside: I think maybe it would make more sense without the word "replaceable". I think it's redundant because a parameter is a variable that helps define some kind of function (in this case a query), so the name of the parameter in the query definition is meant to be replaced by some value, which I think is what this feature does. (I.e. you follow the syntax explained below to define how the parameter gets its value, and then the query runs with that value. )

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,30 @@
+[[osquery-placeholder-fields]]
+== Use placeholder fields in Osquery queries 
+
+Instead of hard-coding alert and event values into Osquery queries, you can use placeholder fields to dynamically pass this data into queries. Placeholder fields function like replaceable parameters. You can use placeholder fields to build flexible and re-usable queries. 
+
+Placeholder fields work in single queries or query packs. They are also supported in the following features:
+
+* <<alerts-run-osquery,Live queries>>
+* <<osquery-response-action,Osquery Response Action>>
```
benironside, line 9:
```suggestion
* <<osquery-response-action,Osquery Response Actions>>
```

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,30 @@
...
++
+NOTE: Placeholder fields within an Osquery Response Action query will only work if the rule has generated alerts.
```
benironside, line 11: Not sure I understand the purpose of this note. I could be missing something, but why would someone expect a response action to run if it's parent rule hadn't generated an alert?

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+* <<invest-guide-run-osquery,Investigation guides using Osquery queries>> 
+
+[float]
+[[placeholder-field-syntax]]
+=== Placeholder field syntax and requirements
+
+Placeholder fields use http://mustache.github.io/[mustache syntax] and must be wrapped in double curly brackets (`{{}}`). You can use any field within an event or alert document as a placeholder field. 
+
+Queries with placeholder fields _must_ be run against an alert or event. If a placeholder field can't find an alert or event value, the query status will be `Skipped`.
```
benironside, line 20:
```suggestion
Queries with placeholder fields can only run against alerts or events. Otherwise, they will lack the necessary values and be `Skipped`. 
```

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+
+[float]
+[[placeholder-field-example]]
+==== Example query with a placeholder field
+
+In the following query, the `{{host.name}}` placeholder field is used: 
+
+`SELECT * FROM os_version WHERE name = `{{host.name}}``
+
+When you run the query, the value that's stored in the alert or event's `host.name` field will be passed into the `{{host.name}}` placeholder field.
```
benironside, line 30: very clear example, great illustration

docs/osquery/osquery-response-action.asciidoc
```diff
@@ -27,6 +27,9 @@ You can add Osquery Response Actions to new or existing custom query rules. Quer
 ** *Existing rule*: Edit the rule's settings, then go to the *Actions* tab. In the tab, click the *osquery* icon under the Response Actions section.
 . Specify whether you want to set up a single live query or a pack:
 ** *Query*: Select a saved query or enter a new one. After you enter the query, you can expand the **Advanced** section to view or set {kibana-ref}/osquery.html#osquery-map-fields[mapped ECS fields] included in the results from the live query. Mapping ECS fields is optional.
++
+TIP: Use <<osquery-placeholder-fields,placeholder fields>> to dynamically add existing alert data to your query.
```
benironside, line 31:
```suggestion
TIP: You can use <<osquery-placeholder-fields,placeholder fields>> to dynamically add alert data to your query. 
```

benironside (commented):
Looks excellent overall, good examples and I think I could use this feature no problem after reading this — I left a few small suggestions for you to consider

docs/osquery/osquery-placeholder-fields.asciidoc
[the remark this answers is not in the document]
nastasha-solomon:
```suggestion
Instead of hard-coding alert and event values into Osquery queries, you can use placeholder fields to dynamically pass this data into queries. Placeholder fields function like parameters. You can use placeholder fields to build flexible and re-usable queries. 
```

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,29 @@
+[[osquery-placeholder-fields]]
+== Use placeholder fields in Osquery queries 
+
+Instead of hard-coding alert and event values into Osquery queries, you can use placeholder fields to dynamically pass this data into queries. Placeholder fields function like parameters. You can use placeholder fields to build flexible and re-usable queries.
```
jmikell821, line 4:
```suggestion
Instead of hard-coding alert and event values into Osquery queries, you can use placeholder fields to dynamically pass this data into queries. Placeholder fields function like parameters. You can use placeholder fields to build flexible and reusable queries. 
```

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+
+[float]
+[[placeholder-field-example]]
+==== Example query with a placeholder field
+
+The following query uses the `{{host.name}}` placeholder field: 
+
+`SELECT * FROM os_version WHERE name = `{{host.name}}``
+
+When you run the query, the value that's stored in the alert or event's `host.name` field will be passed into the `{{host.name}}` placeholder field.
```
jmikell821, line 29: `passed into` sounds a little strange. Would `transferred` work here? 
```suggestion
When you run the query, the value that's stored in the alert or event's `host.name` field will be transferred to the `{{host.name}}` placeholder field. 
```

jmikell821 (commented):
Two small comments, otherwise, looks good!

docs/osquery/osquery-placeholder-fields.asciidoc
nastasha-solomon: @benironside I'm fairly certain that the second sentence needs to be kept as is. If an Osquery query is using a placeholder field to retrieve a field that doesn't exist on the alert or event, the query will fail regardless of where it was ran against. @szwarckonrad @tomsonpl can you verify what I said is correct? 

@szwarckonrad @tomsonpl - On a related note, during my testing, I've seen the `error` status returned for failed queries (i.e. queries using placeholder fields to pass in values for fields that don't exist on an alert or event). Is this correct? Should I replace `Skipped` with `error`?

docs/osquery/osquery-placeholder-fields.asciidoc
nastasha-solomon:
```suggestion
When you run the query, the value that's stored in the alert or event's `host.name` field will be passed into the `{{host.os.name}}` placeholder field. 
```

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+
+[float]
+[[placeholder-field-example]]
+==== Example query with a placeholder field
+
+The following query uses the `{{host.name}}` placeholder field: 
+
+`SELECT * FROM os_version WHERE name = `{{host.name}}``
+
+When you run the query, the value that's stored in the alert or event's `host.name` field will be passed into the `{{host.os.name}}` placeholder field.
```
nastasha-solomon, line 29:
```suggestion
When you run the query, the value that's stored in the alert or event's `host.name` field will be transferred to the `{{host.os.name}}` placeholder field. 
```

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+
+Queries with placeholder fields _must_ be run against an alert or event. If a placeholder field can't find an alert or event value, the query status will be `Skipped`. 
+
+[float]
+[[placeholder-field-example]]
+==== Example query with a placeholder field
+
+The following query uses the `{{host.name}}` placeholder field: 
+
+`SELECT * FROM os_version WHERE name = `{{host.name}}``
```
nastasha-solomon, line 27:
```suggestion
`SELECT * FROM os_version WHERE name = `{{host.os.name}}``
```

docs/osquery/osquery-placeholder-fields.asciidoc
tomsonpl: Hey @nastasha-solomon , @benironside - to me both of your version seem the same (correct) :P 

Whenever we use the {{ data }} dynamic params syntax within context of Alert or Event (we need its data) - we try to replace the parameter with the data, and if we fail (eg. there is no data held under the key we provided) - we return an `error` saying "This query hasn't been called due to parameter used and its value not found in the alert." which means that in the end we didn't run this osquery on the host (not sending action to fleet). To me this can safely be referred to as something like "The query was skipped" 

I am sorry as this might sound confusing, but I'll be happy to discuss this further if that would help 👍

nastasha-solomon: Ah, ok thanks for the additional information about the error message and how Fleet perceives (or doesn't perceive?) the "failed" query. With this in mind, I agree that @benironside's edit makes sense to use. Thanks, both!

docs/osquery/osquery-placeholder-fields.asciidoc
```diff
@@ -0,0 +1,29 @@
...
++
+* <<invest-guide-run-osquery,Investigation guides using Osquery queries>> 
+
+[float]
+[[placeholder-field-syntax]]
+=== Placeholder field syntax and requirements
+
+Placeholder fields use http://mustache.github.io/[mustache syntax] and must be wrapped in double curly brackets (`{{example.field}}`). You can use any field within an event or alert document as a placeholder field. 
+
+Queries with placeholder fields can only run against alerts or events. Otherwise, they will lack the necessary values and be `Skipped`.
```
nastasha-solomon, line 19:
```suggestion
Queries with placeholder fields can only run against alerts or events. Otherwise, they will lack the necessary values and the query status will be `error`. 
```

Issue #1181, opened 2021-10-28
[DOCS] Document privileges needed for CCS rules


rylnd:
## Description 

We recently had an [SDH](https://github.com/elastic/sdh-security-team/issues/229) where the underlying issue was related to CCS and API keys. The behavior is described [here](https://elastic.slack.com/archives/CA6N3HDF1/p1635404144037000), but the TLDR is that:

* CCS with API keys is different than CCS for users
* remote privileges don't matter; local privileges are applied to the remote
* local privileges are captured when the API key is created (i.e. a snapshot)

## Acceptance Test Criteria

* User can correctly configure CCS permissions for their detection rules

Related: https://github.com/elastic/security-team/issues/1983

rylnd:
It doesn't look like there are [existing docs](https://github.com/elastic/elasticsearch/issues/72841) to reference from the elasticsearch side, but I'm happy to help iterate on this one.

joepeeples:
Merged #3119.

PR #3119, opened 2023-03-31
Cross-cluster search and detection rules

### LINKED ISSUE #3033: Detections rules in cross cluster search
## Description

There were number of issues related to set up of detections rules to work in cross cluster envronment
- https://github.com/elastic/sdh-security-team/issues/520
- https://github.com/elastic/sdh-security-team/issues/555

The troubles seemed to be rooted in role configuration on the local and remote clusters.
There are some differences, how role applied to remote cluster depends on the way rule is executed:
 - when rule runs in preview, it uses basic authorization
 - during actual scheduled rule execution, it uses API token, which is a snapshot of the user privileges, on the time of API key creation

I didn't find any page on Security Documentation that gives an example of a rule set up in CCS.
Existing pages in ES documentation doesn't cover Security Solution specific nuances,  [for reference](https://www.elastic.co/guide/en/elasticsearch/reference/8.6/remote-clusters-privileges.html#clusters-privileges-ccs-kibana)
I think it's might be beneficial to have one, with few of the examples and which highlights potential issues.

Based on the mentioned SDH tickets, I would focus document on:

- role on local cluster should give the same privileges as role on the remote cluster
  This due to the way how security authorization model works
  > CCS with users: an authenticated user in your local cluster is granted the same roles on the remote cluster. However, you have to make sure you have the same roles defined in your local cluster and your remote cluster for this to work as expected. If you don't define the same role on the remote cluster, the user won't have privileges on the remote cluster.
  > CCS with API keys: your API key is granted the same privileges on the remote cluster, regardless of whether the source roles are defined on the remote cluster or not
   - [reference 1](https://github.com/elastic/sdh-security-team/issues/555#issuecomment-1451877679)
   - [reference 2](https://github.com/elastic/sdh-security-team/issues/520#issuecomment-1406698489)
- after change in role on the local cluster, API key for rule should be [regenerated](https://github.com/elastic/sdh-security-team/issues/555#issuecomment-1452023792)
- warning https://www.elastic.co/guide/en/security/8.6/advanced-settings.html#enable-ccs-warning, can be avoided by [adding remote cluster prefix with index name to list of indices in the role](https://github.com/elastic/sdh-security-team/issues/520#issuecomment-1405074753)

## Acceptance Test Criteria

Documentation page or section with the examples how to set up rule to work in CCS environment
 

## Notes

- Add all appropriate labels to the issue, especially the version number label.
- Be sure to add any necessary screenshots, code text or console commands for clarity.
- Include any conditions or caveats that may affect customers.

joepeeples:
Resolves #3033, #1181.

Preview: [Cross-cluster search and detection rules](https://security-docs_3119.docs-preview.app.elstc.co/guide/en/security/master/rules-cross-cluster-search.html)


rylnd (approved):
This looks fantastic @joepeeples, thanks for the diligence here. I would agree that this does close https://github.com/elastic/security-docs/issues/1181 as well.

docs/detections/rules-cross-cluster-search.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+.. With this user, <<rules-ui-create,configure a rule>> to search the remote indices. When creating or editing a rule, enter the `<remote_cluster_name>:<index_name>` pattern in the *Source* section.
++
+[role="screenshot"]
+image::images/ccs-rule-source.png[Rule source configuration]
++
+NOTE: If the rule's *Source* uses a data view instead of index patterns, you must also define the data view using the `<remote_cluster_name>:<index_name>` pattern. Refer to {kibana-ref}/data-views.html#management-cross-cluster-search[Use data views with cross-cluster search] for more on defining a data view.
+
+.. (Optional) <<preview-rules,Preview the rule>> to test its expected results.
++
+IMPORTANT: The rule preview uses the currently logged-in user's cross-cluster search privileges, while the rule itself runs using the privileges saved in its API key. The preview results could be different from the rule's actual behavior if the user performing the preview has different privileges from what's saved in the rule's API key.
```
vitaliidm, line 47:
```suggestion
IMPORTANT: The rule preview uses the currently logged-in user's cross-cluster search privileges, while the rule itself runs using the privileges snapshot saved in its API key at the moment of key creation. The preview results could be different from the rule's actual behavior if the user performing the preview has different privileges from what's saved in the rule's API key.
```

vitaliidm (approved):
Thanks a lot @joepeeples. This page should be very helpful both for our customers and devs.

Would it be possible to backport this page to at least 8.6 version, where this behaviour was confirmed?

joepeeples:
> Would it be possible to backport this page to at least 8.6 version, where this behaviour was confirmed?

@vitaliidm I was going to ask about backporting... yes! We could probably go back farther, since I know CCS has been a vexing issue for a while... Would you want to backport _beyond_ 8.6? and how far back?

I think you mentioned that might require additional testing in those earlier versions, so we could break out any additional backports into a separate issue/PR, so we can merge this PR more quickly and at least get the instructions into 8.6+.

docs/detections/rules-cross-cluster-search.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+
+.. Assign the role you just created to a user who you want to configure your cross-cluster detection rules.
++
+[IMPORTANT]
+====
+* This step ensures that the privileges to read remote indices are applied from the user to the rule itself. When a user creates a new rule or saves edits to an existing rule, their current privileges are saved to the rule's API key. If that user’s privileges change in the future, you'll need to <<update-api-key,update the rule's API key>> to update the privileges that the rule uses.
+* This user must also have the <<enable-detections-ui,appropriate privileges>> to manage and preview rules.
+====
+
+.. With this user, <<rules-ui-create,configure a rule>> to search the remote indices. When creating or editing a rule, enter the `<remote_cluster_name>:<index_name>` pattern in the *Source* section.
```
benironside, line 38:
```suggestion
.. As this user, to <<rules-ui-create,configure a rule>> that searches remote indices: create or edit a rule, then enter the `<remote_cluster_name>:<index_name>` pattern in the *Source* section.
```

docs/detections/rules-cross-cluster-search.asciidoc
```diff
@@ -0,0 +1,65 @@
...
++
+[role="screenshot"]
+image::images/ccs-rule-source.png[Rule source configuration]
++
+NOTE: If the rule's *Source* uses a data view instead of index patterns, you must also define the data view using the `<remote_cluster_name>:<index_name>` pattern. Refer to {kibana-ref}/data-views.html#management-cross-cluster-search[Use data views with cross-cluster search] for more on defining a data view.
```
benironside, line 43:
```suggestion
NOTE: If the rule's *Source* uses a data view instead of an index pattern, you still need to define the data view using the `<remote_cluster_name>:<index_name>` pattern. Refer to {kibana-ref}/data-views.html#management-cross-cluster-search[Use data views with cross-cluster search] for more on defining a data view.
```
I could be missing something here but I read this as meaning "If the rule's source uses a data view, use the same pattern you would have used had you not read this note." In which case I wouldn't be sure what the first sentence adds. To me it wasn't obvious after reading it why the user should go to the kibana docs.

docs/detections/rules-cross-cluster-search.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+[role="screenshot"]
+image::images/ccs-remote-role.png[Remote cluster role configuration]
+
+. On the local cluster:
+
+.. Assign the role you just created to a user who you want to configure your cross-cluster detection rules.
++
+[IMPORTANT]
+====
+* This step ensures that the privileges to read remote indices are applied from the user to the rule itself. When a user creates a new rule or saves edits to an existing rule, their current privileges are saved to the rule's API key. If that user’s privileges change in the future, you'll need to <<update-api-key,update the rule's API key>> to update the privileges that the rule uses.
```
benironside, line 34:
```suggestion
* This step ensures that the privileges to read remote indices are applied from the user to the rule itself. When a user creates a new rule or saves edits to an existing rule, their current privileges are assigned to the rule's API key. If that user’s privileges change in the future, the rule's API key will not update until you manually <<update-api-key,update the rule's API key>>.
```

docs/detections/rules-cross-cluster-search.asciidoc
```diff
@@ -0,0 +1,65 @@
...
++
+IMPORTANT: The rule preview uses the currently logged-in user's cross-cluster search privileges, while the rule itself runs using the privileges snapshot saved in its API key at the moment of key creation. The preview results could be different from the rule's actual behavior if the user performing the preview has different privileges from what's saved in the rule's API key.
+
+.. Save and enable the rule.
+
+[discrete]
+[[update-api-key]]
+=== Update a rule's API key
+
+When a user creates a new rule or saves edits to an existing rule, their current privileges are saved to {kibana-ref}/alerting-setup.html#alerting-authorization[the rule's API key]. If that user's privileges change in the future, the rule *_does not_* automatically update with the user's latest privileges — you must update the rule's API key to update the privileges that the rule uses.
```
benironside, line 55:
```suggestion
When a user creates a new rule or saves edits to an existing rule, their current privileges are saved to {kibana-ref}/alerting-setup.html#alerting-authorization[the rule's API key]. If that user's privileges change in the future, the rule *_does not_* automatically update with the user's latest privileges — you must update the rule's API key to update its privileges.
```

docs/detections/rules-cross-cluster-search.asciidoc
joepeeples: Good catch, I think I tried to cram too much into that first sentence. It's meant to say that data sources also require specific configuration to work with cross-cluster search, and that configuration is actually a separate process & UI from defining the rule. Let me take another pass at this...

docs/detections/rules-cross-cluster-search.asciidoc
joepeeples: I'd like to adopt much of your suggestion, but still keep the step as a more direct instruction ("do this" vs. "to do this, do that"). I think that tone is more consistent with most of our other steps. But your your other wording suggestions def make this a snappier step:
```suggestion
.. As this user, <<rules-ui-create,configure a rule>> that searches the remote indices: create or edit a rule, then enter the `<remote_cluster_name>:<index_name>` pattern in the *Source* section.
```

docs/detections/rules-cross-cluster-search.asciidoc
joepeeples: Nit: I think "saved" is more accurate, because the API key isn't assigned a privilege so _it_ can do something, it's more like a piece of data that contains the privilege itself, which _the rule_ uses so it can do something, if that makes any sense. But I dig the revision to the last sentence, slight revision to avoid repeating "the rules' API key":
```suggestion
* This step ensures that the privileges to read remote indices are applied from the user to the rule itself. When a user creates a new rule or saves edits to an existing rule, their current privileges are saved to the rule's API key. If that user’s privileges change in the future, the rule's API key will not update until you manually update it. Refer to <<update-api-key>> for details.
```

docs/detections/rules-cross-cluster-search.asciidoc
joepeeples: @benironside maybe something like this? Would it be better to omit "using the x:y pattern"? I wanted to keep reinforcing the sample pattern throughout, but we could omit if it's misleading or too much in one sentence.
```suggestion
NOTE: If the rule's *Source* uses a data view instead of index patterns, that data view must be defined separately, using the `<remote_cluster_name>:<index_name>` pattern. Refer to {kibana-ref}/data-views.html#management-cross-cluster-search[Use data views with cross-cluster search] for more on defining a data view.
```

nastasha-solomon: Do you think it'd be helpful lead-in to the Kibana doc link to specify that we're talking about a data view with cross-cluster search? For example:

```suggestion
NOTE: If the rule's *Source* uses a data view with cross-cluster search instead, you must define the data view separately, using the `<remote_cluster_name>:<index_name>` pattern. Refer to {kibana-ref}/data-views.html#management-cross-cluster-search[Use data views with cross-cluster search] for more on defining a data view.
```

docs/detections/rules-cross-cluster-search.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+
+This section explains the general process for setting up cross-cluster search in detection rules. For specific instructions on each part of the process, refer to the linked documentation.
+
+. On the local cluster, {ref}/remote-clusters-connect.html[set up a connection] to the remote cluster. Note the unique name that you give to the remote cluster, because you'll need to use it throughout this process.
+
+. On both the local and remote clusters, {ref}/remote-clusters-privileges.html#clusters-privileges-ccs-kibana[create a role for cross-cluster search privileges], and make sure the two roles have _identical_ names. Assign each role the following privileges:
+
+.. *Local cluster role*: Assign the `read` privilege to the indices you want to search, using _both_ the local and remote index patterns for each index. To specify a remote index, use the pattern `<remote_cluster_name>:<index_name>`.
++
+For example, if the remote cluster's name is `remote-security-data` and you want to query the `logs-*` indices, include both `logs-*` and `remote-security-data:logs-*` and assign them the `read` privilege.
```
nastasha-solomon, line 18: Should this be index patterns instead of indices? 

```suggestion
For example, if the remote cluster's name is `remote-security-data` and you want to query the `logs-*` index pattern, include both `logs-*` and `remote-security-data:logs-*` and assign them the `read` privilege.
```

I feel like both might work in this case, but one might be more technically correct than the other. I also hesitate to suggest this change because I understand that you're saying "if you want to query all of the log indices" or "all of the indices that begin with `logs-`" -- so basically it works as is. Wdyt?

docs/detections/rules-cross-cluster-search.asciidoc
```diff
@@ -0,0 +1,65 @@
+[[rules-cross-cluster-search]]
+== Cross-cluster search and detection rules
```
nastasha-solomon, line 2: What are your thoughts on changing this to something that's more Security-focused, for example: 

```suggestion
== Detect threats across clusters
```

Another option that's more search-focused is:

```suggestion
== Search for threats across clusters
```

The current title is already clear and descriptive, so these are just suggestions :)

nastasha-solomon (approved):
This looks great, @joepeeples. Good job on keeping it focused and streamlined!

docs/detections/rules-cross-cluster-search.asciidoc
```diff
@@ -0,0 +1,65 @@
...
+.. As this user, <<rules-ui-create,configure a rule>> that searches the remote indices: create or edit a rule, then enter the `<remote_cluster_name>:<index_name>` pattern in the *Source* section.
++
+[role="screenshot"]
+image::images/ccs-rule-source.png[Rule source configuration]
++
+NOTE: If the rule's *Source* uses a data view instead of index patterns, you must also define the data view using the `<remote_cluster_name>:<index_name>` pattern. Refer to {kibana-ref}/data-views.html#management-cross-cluster-search[Use data views with cross-cluster search] for more on defining a data view.
+
+.. (Optional) <<preview-rules,Preview the rule>> to test its expected results.
++
+IMPORTANT: The rule preview uses the currently logged-in user's cross-cluster search privileges, while the rule itself runs using the privileges snapshot saved in its API key at the moment of key creation. The preview results could be different from the rule's actual behavior if the user performing the preview has different privileges from what's saved in the rule's API key.
```
jmikell821, line 47:
```suggestion
IMPORTANT: The rule preview uses the current user's cross-cluster search privileges, while the rule itself runs using the privileges snapshot saved in its API key the moment the key is created. The preview results could be different from the rule's actual behavior if the user performing the preview has different privileges than what's saved in the rule's API key.
```

jmikell821 (commented):
Just a couple of small suggestions in addition to the rest of the great feedback already left. Looks great, thanks!

docs/detections/rules-cross-cluster-search.asciidoc
joepeeples: @nastasha-solomon I'd like to keep "uses a data view instead of index patterns" bc it directly references the UI options, but I think we could do something similar to what you suggested:
```suggestion
NOTE: If the rule's *Source* uses a data view instead of index patterns, you must define the data view for cross-cluster search separately, using the `<remote_cluster_name>:<index_name>` pattern. Refer to {kibana-ref}/data-views.html#management-cross-cluster-search[Use data views with cross-cluster search] for more on defining a data view.
```

Issue #3136, opened 2023-04-05
8.7.1 Release notes 


nastasha-solomon:
### Description
Generate the 8.7.1 release notes and publish them before or on the release day, which is **February 21, 2023**.

**Milestone dates:**
- **FF:** April 11, 2023
- **BC:** April 12, 2023
- **Release:** April 18, 2023

### Steps
1. Create a new branch using GH Desktop or the CLI. Here are the steps for the CLI: 
   a. From the CLI, run `git checkout main` to checkout the `main` branch.
   b. Run `git fetch origin` to retrieve the latest version of the `main` branch.
   c. Run `git pull` to pull down the latest changes in the `main` branch.
   d. Run `git checkout -b issue-XXX-8.7.1-rn` to create a new branch.
2. Open the Security docs project in your text editor (VS Code).
3. Expand the `release-notes` folder the open the `8.7.asciidoc` file.
4. At the top of the file, create a new section for the `8.7.1` release notes, for example:

``` 
[discrete]
[[release-notes-8.7.1]]
=== 8.7.1

[discrete]
[[bug-fixes-8.7.1]]
==== Bug fixes and enhancements
* TBD
``` 

5. Open the `release-notes.asciidoc` file and insert a link to the `8.7.1` release notes, for example: `<<release-notes-8.7.1, {elastic-sec} version 8.7.1>>`
6. Save your changes.
7. Commit and push your changes:
   a. Run `git add --all` to add changes from all tracked and untracked files.
   b. Run `git commit -m "A brief description of your change"` to commit your changes. 
   c. Run `git push` to push your changes to the remote branch. If the branch doesn’t already exist in the remote repo, it is automatically created.
8. Open a PR in GH. Remember to add the `8.7` label so the release notes changes are backported to the 8.7 branch after you merge them into `main`.  

### Notes:
- If you add the PR to the Endpoint feature/bug/known issue, use the hardcoded URL to create the reference link. [Here](https://www.elastic.co/guide/en/security/8.3/release-notes-header-8.1.0.html#bug-fixes-8.1.2) is an example.
- ResponseOps features for cases, connectors, and alerts are always doc'd in the Kibana release notes _unless_ they are Security-specific features. This is also the case for Osquery feature and enhancements. 
- Unless requested, we typically do not need to update the breaking changes section for patch releases.
[...]

PR #3137, opened 2023-04-05
8.7.1 Release notes 

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/3136.

Previews:
- [Index page](https://security-docs_3137.docs-preview.app.elstc.co/guide/en/security/master/release-notes.html)
- [Release notes](https://security-docs_3137.docs-preview.app.elstc.co/guide/en/security/master/release-notes-header-8.7.0.html#release-notes-8.7.1)


docs/release-notes/8.7.asciidoc
```diff
@@ -1,6 +1,21 @@
...
+[discrete]
+[[enhancements-8.7.1]]
+==== Enhancements
+There are no user-facing changes in 8.7.1.
+
+[discrete]
+[[bug-fixes-8.7.1]]
+==== Bug fixes
+* Fixes a UI bug that affected the rule details page when rules had long queries ({pull}153338[#153338]).
+* Fixes a bug that stopped Elastic Defend from working when Malware protection was enabled on Linux hosts.
```
joepeeples, line 17:
```suggestion
* Fixes a bug that stopped {elastic-defend} from working when malware protection was enabled on Linux hosts.
```

joepeeples (commented):
Small edit, otherwise LGTM!

roxana-gheorghe (approved):

docs/release-notes/8.7.asciidoc
```diff
@@ -1,6 +1,22 @@
...
+
+[discrete]
+[[enhancements-8.7.1]]
+==== Enhancements
+There are no user-facing changes in 8.7.1.
+
+[discrete]
+[[bug-fixes-8.7.1]]
+==== Bug fixes
+* Fixes a bug that caused the {agent} upgrade from 8.6.2 to 8.7.0 to fail for {agents} running the Osquery Manager integration (https://github.com/elastic/elastic-agent/pull/2448[#2448]).
```
kevinlog, line 16: we should note here that upgrade `8.7.0` -> `8.7.1` could still cause an upgrade issue.  We should include the same workarounds as below.  After the upgrade to `8.7.1`, the problem will be fixed.

I'm referring to this section from 8.7.0 known issues.

```
The {agent} upgrade from 8.6.2 to 8.7.0 might fail for {agents} running the Osquery Manager integration (https://github.com/elastic/elastic-agent/issues/2433[#2433]). To prevent this, delete the Osquery Manager integration from your {agents} before upgrading them to 8.7.0, then re-add it after upgrading
```

docs/release-notes/8.7.asciidoc
```diff
@@ -1,6 +1,53 @@
...
+
++
+
+[NOTE]
+
+=====
+If you are upgrading {agent} to 8.7.1 from 8.6.2 to 8.7.0, the {agent} upgrade might fail for {agents} running the Osquery Manager integration. To prevent this, delete the Osquery Manager integration from your {agents} before upgrading them to 8.7.0, then re-add it after upgrading.
```
joepeeples, line 23: This seems to contradict the previous item, which says that we already fixed this bug. So why is the same behavior still happening? Is the idea that we _did_ fix the bug, but users might still have to delete and re-add the integration to get the fix? If that's the case, maybe rephrase this so it doesn't sound identical to the bug fix above. 

Also, should the last sentence be "before upgrading them to 8.7.1" since this is the 8.7.1 section?

docs/release-notes/8.7.asciidoc
```diff
@@ -1,6 +1,53 @@
...
+=====
+If the error message `Failed to download or validate user artifacts` appears in the Policy Response section of your Elastic Defend integration policy, you can resolve the error by adding the affected endpoint policy artifacts as <<trusted-apps-ov,global trusted applications>>, <<endpoint-rule-exceptions,endpoint exceptions>>, <<event-filters,event filters>>, <<blocklist,blocklist entries>>, or <<host-isolation-exceptions,host isolation exceptions>>. 
+
+
+To find the affected endpoint policy artifacts:
+
+. Go to *Fleet -> Agent policies*.
+. Open the impacted {agent} policy.
+. Select *Actions -> View policy*.
+. Search for `compression_algorithm: none` in the policy details. This field and value pair will appear under the affected endpoint policy artifacts.
```
joepeeples, line 47:
```suggestion
. Search for `compression_algorithm: none` in the policy details. This field-value pair will appear under the affected endpoint policy artifacts.
```

docs/release-notes/8.7.asciidoc
nastasha-solomon: @joepeeples @kevinlog does the following revision work? It emphasizes that this is only a problem for users on 8.6.2 or 8.7.0, but that it's preventable.

```suggestion
To prevent upgrade issues while upgrading {agent} from 8.6.2 or 8.7.0, delete the Osquery Manager integration from your {agents} before upgrading them, then re-add it after upgrading. 

```

joepeeples: LGTM, thanks!

kevinlog: @nastasha-solomon this looks good to me, thanks!

docs/release-notes/8.7.asciidoc
```diff
@@ -1,6 +1,54 @@
...
+
++
+
+[NOTE]
+
+=====
+To prevent upgrade issues while upgrading {agent} from 8.6.2 or 8.7.0, delete the Osquery Manager integration from your {agents} before upgrading them, then re-add it after upgrading. 
+
+
+If you already upgraded your {agents}, and they're are stuck in the `Updating` state, you'll need to upgrade them using the API instead of {fleet}. First, remove the Osquery integration from the {agents}, then use the https://petstore.swagger.io/?url=https://raw.githubusercontent.com/elastic/kibana/8.7/x-pack/plugins/fleet/common/openapi/bundled.json#/default/upgrade-agent[Agent upgrade API] to start the upgrade. After the upgrade completes, add the Osquery Manager integration to the {agents} again.
```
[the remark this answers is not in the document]
nastasha-solomon: Yes, I believe that's correct. Users will encounter the Agent upgrade issue only if they're upgrading from 8.6.2 or 8.7.0 to a later version.

cc: @kevinlog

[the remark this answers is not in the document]
kevinlog: @nastasha-solomon 

> Yes, I believe that's correct. Users will encounter the Agent upgrade issue only if they're upgrading from 8.6.2 or 8.7.0 to a later version.

This is correct. cc @jmikell821

yctercero (approved):

benironside (approved):
LGTM!

PR #3244, opened 2023-05-09
Response action: `execute`

### LINKED ISSUE #3215: Response action: `execute`
## Description 

Related: 
* Epic: https://github.com/elastic/security-team/issues/5441
   Epic includes 3 response actions (`upload`, `execute`, `execute-file`), but only `execute` is releasing in 8.8
* Issue: https://github.com/elastic/security-team/issues/5370

## Notes
* UI help text also under review: https://github.com/elastic/security-team/issues/6575

joepeeples:
Resolves #3215.

Previews:
- [Elastic Defend feature privileges](https://security-docs_3244.docs-preview.app.elstc.co/guide/en/security/master/endpoint-management-req.html) -- added to end of table
- [Endpoint response actions | `execute`](https://security-docs_3244.docs-preview.app.elstc.co/guide/en/security/master/response-actions.html#_execute) -- new section


docs/management/admin/response-actions.asciidoc
```diff
@@ -111,6 +111,29 @@ TIP: You can use the <<use-osquery,Osquery manager integration>> to query a host
 When {elastic-defend} prevents file activity due to <<malware-protection,malware prevention>>, the file is quarantined on the host and a malware prevention alert is created. To retrieve this file with `get-file`, copy the path from the alert's *Quarantined file path* field (`file.Ext.quarantine_path`), which appears under *Highlighted fields* in the alert details flyout. Then paste the value into the `--path` parameter. 
 ====
 
+=== `execute`
+
+Run a shell command on the host. The command's output and any errors are returned in the response console, up to 2000 characters. The complete output (stdout and stderr) are also saved to a downloadable .zip archive (password: `elastic`). Use these parameters:
+
+* `--command` : (Required) A single-line command to run on the host. The command must be supported by the host operating system's default terminal shell (for example, bash for Linux, cmd.exe for Windows).
```
benironside, line 118: I'm not clear on what single-line means here. Is there a character limit? Would it be accurate to just say "A shell command to run on the host"?

docs/management/admin/response-actions.asciidoc
```diff
@@ -111,6 +111,29 @@ TIP: You can use the <<use-osquery,Osquery manager integration>> to query a host
...
++
+[NOTE]
+====
+* Multiple consecutive dashes in the value must be escaped. For example, to represent a directory named `/opt/directory--name`, use the following: `/opt/directory\-\-name`.
```
benironside, line 122: Interesting that dashes only need to be escaped if there are two, but in that case _both_ need to be escaped (not just the second one). Might be worth making explicit whether the user has to escape a single dash.

docs/management/admin/response-actions.asciidoc
```diff
@@ -111,6 +111,29 @@ TIP: You can use the <<use-osquery,Osquery manager integration>> to query a host
...
+
+====
+
+* `--timeout` : (Optional) How long the host should wait for the command to complete. Use `h` for hours, `m` for minutes, `s` for seconds (for example, `2s` is two seconds). If no timeout is given, it defaults to four hours.
+
+Required privilege: *Execute Operations*
+
+Example: `execute --command "ls -al" --timeout 2s --comment "Get list of all files"`
+
+WARNING: This response action runs commands as a highly privileged, root-level user on the host. Be careful with any commands that could cause irrevocable changes.
```
benironside, line 135:
```suggestion
WARNING: This response action runs commands as root. Be careful with any commands that could cause irrevocable changes.
```
If this suggestion is accurate I think the same comment from above applies.

benironside (commented):
Looks good, this feature is so cool. Left some suggestions for your consideration, lmk if you have any questions.

docs/management/admin/response-actions.asciidoc
```diff
@@ -111,6 +111,29 @@ TIP: You can use the <<use-osquery,Osquery manager integration>> to query a host
 When {elastic-defend} prevents file activity due to <<malware-protection,malware prevention>>, the file is quarantined on the host and a malware prevention alert is created. To retrieve this file with `get-file`, copy the path from the alert's *Quarantined file path* field (`file.Ext.quarantine_path`), which appears under *Highlighted fields* in the alert details flyout. Then paste the value into the `--path` parameter. 
 ====
 
+=== `execute`
+
+Run a shell command on the host. The command's output and any errors are returned in the response console, up to 2000 characters. The complete output (stdout and stderr) are also saved to a downloadable .zip archive (password: `elastic`). Use these parameters:
```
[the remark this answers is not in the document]
joepeeples: This says it's downloadable, so it would be like any other file you download using a browser. It's also probably pretty clear to the user actually performing the actions, because the browser will either ask the user where to save it, or save it to the Downloads folder by default.

docs/management/admin/response-actions.asciidoc
joepeeples: That might work, but I'm not certain. This is meant to say that the user can't run commands that require multiple lines (or returns, newlines, etc.?). Everything you want to do has to be on a single command line: either a single command, or multiple commands piped or concatenated or otherwise combined on a single line. I was trying to convey that as concisely as possible, without too much exposition.

@paul-tavares @kevinlog WDYT?

docs/getting-started/defend-feature-privs.asciidoc
```diff
@@ -47,6 +47,11 @@ To grant access, select *All* for the *Security* feature in the *{kib} privilege
 
 | *File Operations*
 | Perform file-related <<response-actions,response actions>> in the response console.
+
+| *Execute Operations*
+a| Perform shell commands and script-related <<response-actions,response actions>> in the response console.
+
+WARNING: This allows a user to run highly privileged commands on a host as a root-level user. Restrict this feature privilege to only {elastic-sec} users who require this level of access.
```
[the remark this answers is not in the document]
paul-tavares: The command will be run with as the user that is also running the Endpoint on the host. That normally is a highly privileged system account - `root` likely on Linux/Mac and something like it on Windows. 

Maybe something like:

> WARNING: The commands are executed on the host using the same user account that runs the Elastic Defend integration, which is normally an account that has full control over the system. Restrict this feature privilege to only {elastic-sec} users who require this level of access.

docs/management/admin/response-actions.asciidoc
paul-tavares: I'm ok with either the current value or the suggestion by @benironside .

The thing I would probably remove is "default terminal shell" and instead just mention `bash` for linux and Mac, and `cmd.exe` for Windows. Thats because the default terminal currently on Mac is `zsh`

joepeeples: @paul-tavares Are you saying that this response action will use bash on a macOS system, even if its default is actually zsh?

I was thinking "default terminal shell" could help future-proof in case one of these systems changes their default, like macOS did a few years ago when they went from bash to zsh.

paul-tavares: Correct. that is my understanding. We had decided to use specific shells since we really can't control what the user (the user id that the command is executed as) has setup as their default shell.

docs/management/admin/response-actions.asciidoc
joepeeples: If we're going to the trouble of mentioning that consecutive dashes require escaping, the reader would probably infer that single dashes _don't_ require it -- or else we would have mentioned them, too. (We do mention quote marks below, but that's kind of a special case, and it's a slippery slope to identify the characters that _don't_ require special handling.)

I tinkered with making this explicit anyway, but it started feeling a bit overwrought and distracting from the main point about the consecutive dashes.

docs/getting-started/defend-feature-privs.asciidoc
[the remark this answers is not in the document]
joepeeples:
```suggestion
WARNING: The commands are run on the host using the same user account running the {elastic-defend} integration, which normally has full control over the system. Restrict this feature privilege to only {elastic-sec} users who require this level of access.
```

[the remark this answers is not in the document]
paul-tavares: LGTM  😄

docs/management/admin/response-actions.asciidoc
joepeeples:
```suggestion
* `--command` : (Required) A shell command to run on the host. The command must be supported by `bash` for Linux and macOS hosts, and `cmd.exe` for Windows).
```

paul-tavares: LGTM 👍

docs/management/admin/response-actions.asciidoc
```diff
@@ -111,6 +111,29 @@ TIP: You can use the <<use-osquery,Osquery manager integration>> to query a host
...
+
+* `--command` : (Required) A shell command to run on the host. The command must be supported by `bash` for Linux and macOS hosts, and `cmd.exe` for Windows).
```
joepeeples, line 118:
```suggestion
* `--command` : (Required) A shell command to run on the host. The command must be supported by `bash` for Linux and macOS hosts, and `cmd.exe` for Windows.
```

paul-tavares (approved):
👍

docs/management/admin/response-actions.asciidoc
```diff
@@ -111,6 +111,29 @@ TIP: You can use the <<use-osquery,Osquery manager integration>> to query a host
...
+[NOTE]
+====
+* Multiple consecutive dashes in the value must be escaped. For example, to represent a directory named `/opt/directory--name`, use the following: `/opt/directory\-\-name`.
+
+* You can use quotation marks without escaping. For example: +
+`execute --command "cd "C:\Program Files\directory""`
+
+====
+
+* `--timeout` : (Optional) How long the host should wait for the command to complete. Use `h` for hours, `m` for minutes, `s` for seconds (for example, `2s` is two seconds). If no timeout is given, it defaults to four hours.
```
jmikell821, line 129:
```suggestion
* `--timeout` : (Optional) How long the host should wait for the command to complete. Use `h` for hours, `m` for minutes, `s` for seconds (for example, `2s` is two seconds). If no timeout is specified, it defaults to four hours.
```

jmikell821 (commented):
Looks great! Just a few small nits, thanks!

docs/management/admin/response-actions.asciidoc
kevinlog: > I tinkered with making this explicit anyway, but it started feeling a bit overwrought and distracting from the main point about the consecutive dashes.

I agree here, I do think having an explicit callout about consecutive dashes is enough and users would understand.  At the most, maybe one sentence at the end would drive the point home - "Single dashes do not need to be escaped."

kevinlog (approved):

docs/getting-started/defend-feature-privs.asciidoc
```diff
@@ -47,6 +47,11 @@ To grant access, select *All* for the *Security* feature in the *{kib} privilege
 
 | *File Operations*
 | Perform file-related <<response-actions,response actions>> in the response console.
+
+| *Execute Operations*
+a| Perform shell commands and script-related <<response-actions,response actions>> in the response console.
+
+WARNING: The commands are run on the host using the same user account running the {elastic-defend} integration, which normally has full control over the system. Restrict this feature privilege to only {elastic-sec} users who require this level of access.
```
nastasha-solomon, line 54: Made some slight modifications to the second sentence to hopefully make it more straightforward. 

```suggestion
WARNING: The commands are run on the host using the same user account running the {elastic-defend} integration, which normally has full control over the system. Only grant this feature privilege to {elastic-sec} users who require this level of access.
```

docs/management/admin/response-actions.asciidoc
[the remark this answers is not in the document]
nastasha-solomon:
```suggestion
Run a shell command on the host. The command's output and any errors are returned in the response console, up to 2000 characters. The complete output (stdout and stderr) are also saved to a downloadable `.zip` archive (password: `elastic`). Use these parameters:
```

nastasha-solomon (approved):
Don't have much to add on top of the existing feedback - LGTM!

docs/management/admin/response-actions.asciidoc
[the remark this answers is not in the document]
joepeeples:
```suggestion
Run a shell command on the host. The command's output and any errors appear in the response console, up to 2000 characters. The complete output (stdout and stderr) are also saved to a downloadable `.zip` archive (password: `elastic`). Use these parameters:
```

docs/management/admin/response-actions.asciidoc
joepeeples:
```suggestion
* Multiple consecutive dashes in the value must be escaped; single dashes do not need to be escaped. For example, to represent a directory named `/opt/directory--name`, use the following: `/opt/directory\-\-name`.
```

docs/management/admin/response-actions.asciidoc
```diff
@@ -111,6 +111,29 @@ TIP: You can use the <<use-osquery,Osquery manager integration>> to query a host
...
+
+Required privilege: *Execute Operations*
+
+Example: `execute --command "ls -al" --timeout 2s --comment "Get list of all files"`
+
+WARNING: This response action runs commands on the host using the same user account running the {elastic-defend} integration, which normally has full control over the system. Be careful with any commands that could cause irrevocable changes.
```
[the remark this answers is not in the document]
joepeeples: I think this changes the meaning somewhat -- this edit says that the host is using the account, but the host doesn't really _do_ anything, it's just the place/resource where things happen. We might just be able to omit "on the host" to make it clearer that the response action is using the account, not the host.

> Also, do we need to specify the integration here -- or just Elastic Defend as a whole?

Not sure what you mean; `Elastic Defend` _**is**_ the integration. Did you mean `Elastic Endpoint`? Because it might be more accurate to say that the user account is running Endpoint. @caitlinbetz @kevinlog Is that correct?

```suggestion
WARNING: This response action runs commands using the same user account that is running {elastic-endpoint}, which normally has full control over the system. Be careful with any commands that could cause irrevocable changes.
```

caitlinbetz (approved):

Issue #3181, opened 2023-04-21
[DOCS] Document the new Maintenance Windows feature


nikitaindik:
**Related issue:** https://github.com/elastic/kibana/issues/155099
**Related PR:** https://github.com/elastic/kibana/pull/155386

## Summary
Explain what Maintenance Windows feature is to Security Solution users. It is a new feature that is going to be released in v8.8. It allows a user to set up time periods during which rule notification actions won't get triggered (rules would still run and generate alerts though). Also user would see a callout on the Rule Management page while a maintenance window is running.

<img width="1260" alt="Screenshot 2023-04-21 at 13 24 11" src="https://user-images.githubusercontent.com/15949146/233655467-f3781a00-ff6a-40e5-9896-bcac26c30f5b.png">

You can view and create maintenance windows on the Stack Management page -> Maintenance Windows (under Alerts and Insights).

<img width="1503" alt="Screenshot 2023-04-21 at 16 04 51" src="https://user-images.githubusercontent.com/15949146/233656677-63795364-5448-4a85-ae67-b3cacc9de1af.png">

Maintenance windows can be set up to run once or repeated based on a schedule.

<img width="1501" alt="Screenshot 2023-04-21 at 16 05 29" src="https://user-images.githubusercontent.com/15949146/233656717-0560542c-b03e-4f82-91b8-bb395349c0e2.png">

nastasha-solomon:
As per https://github.com/elastic/kibana/issues/152032, the Kibana docs will likely be added to the [Manage rules](https://www.elastic.co/guide/en/kibana/master/create-and-manage-rules.html) section. I'll link to that section in the Security docs.

nastasha-solomon:
Note to self to link to kibana docs once they're merged: https://github.com/elastic/kibana/pull/157806

PR #3253, opened 2023-05-10
Document the new Maintenance Windows feature

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/3181.

Preview [here](https://security-docs_3253.docs-preview.app.elstc.co/guide/en/security/master/rules-ui-management.html#edit-rules-settings).


docs/detections/rules-ui-manage.asciidoc
```diff
@@ -139,6 +139,8 @@ Similarly, rules will be skipped if they can't be modified by a bulk edit. For e
 ** *Add rule actions*: Add <<rule-notifications,notification actions>> on all selected rules.
 +
 NOTE: The action frequency you select applies to all actions (both new and existing) on all selected rules. If you don't want to change the frequency of existing actions, update the rules separately.
++
+NOTE: Rule actions that you add during a {ref}/maintenance-windows.html[maintenance window] won't run. They'll resume running after the maintenance window ends.
```
joepeeples, line 143: Will `{kibana-ref}` fix the cross-docs link? Also, I wonder if this should be a little more general -- I know it's in the context of creating a rule, adding an action, but it might be more accurate to say that _any_ rule actions won't run during a window, not just the ones that you add _during_ a window.
```suggestion
NOTE: Rule actions won't run during a {kibana-ref}/maintenance-windows.html[maintenance window]. They'll resume running after the maintenance window ends.
```

nastasha-solomon: Good point and the ref update should fix the angery link.

nikitaindik:
LGTM

nikitaindik (approved):

PR #3175, opened 2023-04-19
Configure and install ==> Install and configure

essodjolo:

PR #3255, opened 2023-05-11
Install / Configure — Minor change

### LINKED ISSUE #3212: Change Elastic Defend docs from "Configure and install" to "Install and Configure"
Topic: [Configure and install the Elastic Defend integration](https://www.elastic.co/guide/en/security/current/install-endpoint.html)

Logically, it makes more sense to swap the verbs because an analyst would first install the integration, then configure it.

benironside:
Fixes #3212

### DIFF
--- a/docs/getting-started/install-endpoint.asciidoc
+++ b/docs/getting-started/install-endpoint.asciidoc
@@ -1,5 +1,5 @@
 [[install-endpoint]]
-= Configure and install the {elastic-defend} integration
+= Install and configure the {elastic-defend} integration
 
 Like other Elastic integrations, {elastic-defend} is integrated into the {agent} using {fleet-guide}/fleet-overview.html[{fleet}]. Upon configuration, the integration allows the {agent} to monitor events on your host and send data to the {security-app}.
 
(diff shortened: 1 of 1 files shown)


joepeeples (approved):
LGTM!

@jmikell821 Is there any need to backport this? The [original request](https://github.com/elastic/security-docs/pull/3175) was based in 7.17 for some reason, but I don't think a backport is needed, definitely not that far back.

jmikell821:
> 

@joepeeples I think we should at least backport this to at least `8.4`, if not all the `8x` series. WDYT?

joepeeples:
> I think we should at least backport this to at least `8.4`, if not all the `8x` series. WDYT?

@jmikell821 No strong preference, but it's a small enough change that if we're backporting, might as well do the whole 8.x series. The only caveat is the integration's name and attribute changed twice since 8.0, so we'd have to manually adjust that. We could still have Mergify create all the backport PRs, but they'll all have `{elastic-defend}` in the title so we can just manually fix them to use the correct attributes for each version:

* 8.5 and later: `{elastic-defend}` - no change needed
* 8.4 and 8.3: `{endpoint-cloud-sec}`
* 8.0 to 8.2: These versions actually don't even use attributes in the title — they have "Endpoint Security" hard-written, but we could change to use the attribute `{endpoint-sec}`

Issue #3206, opened 2023-04-28
[BUG] Add note that alerts attached to cases are not exported 


nastasha-solomon:
[Docs](https://www.elastic.co/guide/en/security/master/cases-open-manage.html#cases-export-import) don't mention that attached alerts aren't exported with cases, which is a limitation that users should be aware of when exporting a case's related objects. After users import a case that used to have alerts, they'll need to re-add the alerts to the case.

Will need to backport this change to 8.3. 

cc: @jonathan-buttner

nastasha-solomon:
Will need to create a separate PR to add this note to 8.3-8.7.

PR #3310, opened 2023-05-18
[BUG][8.8] Add note that alerts attached to cases are not exported

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/3206.

Preview [here](https://security-docs_3310.docs-preview.app.elstc.co/guide/en/security/master/cases-open-manage.html#cases-export-import): Expanded note about items that are not exported with cases.


lcawl (approved):
LGTM! IMO we don't need to similarly update the Kibana or Observability docs since alert attachments aren't supported yet in Stack Management and Observability docs don't cover the import/export steps.

PR #3431, opened 2023-06-08
Remove 8.8.0 BC tag

nastasha-solomon:
Removes the breaking changes tag from the 8.8.0 breaking changes section in the Security release notes. This ensures the 8.8 [Elastic Security breaking changes](https://www.elastic.co/guide/en/elastic-stack/current/security-breaking-changes.html) page reflects the 8.8.1 breaking changes tagged region in the `8.8.asciidoc` file.

### DIFF
--- a/docs/release-notes/8.8.asciidoc
+++ b/docs/release-notes/8.8.asciidoc
@@ -19,6 +19,8 @@
 // NOTE: The breaking-changes tagged regions are reused in the Elastic Installation and Upgrade Guide. The pull attribute is defined within this snippet so it properly resolves in the output.
 // THIS ALSO MEANS IF YOU USE LINKS HERE, THEY SHOULD BE FULL URLS WITH NO ATTRIBUTES
 
+:pull: https://github.com/elastic/kibana/pull/
+
 There are no breaking changes in 8.8.1.
 
 //end::breaking-changes[]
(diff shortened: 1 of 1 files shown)


docs/release-notes/8.8.asciidoc
```diff
@@ -58,7 +58,6 @@ To view a detailed summary of the latest features and enhancements, check out ou
 [[breaking-changes-8.8.0]]
 ==== Breaking changes
 
-//tag::breaking-changes[]
 // NOTE: The breaking-changes tagged regions are reused in the Elastic Installation and Upgrade Guide. The pull attribute is defined within this snippet so it properly resolves in the output.
 // THIS ALSO MEANS IF YOU USE LINKS HERE, THEY SHOULD BE FULL URLS WITH NO ATTRIBUTES
 :pull: https://github.com/elastic/kibana/pull/
```
joepeeples, line 63: I'd remove these lines too. Since this is no longer a snippet/tagged region, we don't need reminders in this particular region on how a snippet works (we also have them in the actual snippet above).

We also don't need to redefine the `pull` attribute here since this is no longer a snippet. But we might want to move the `pull` line to the snippet above -- it's not technically necessary because there aren't any links in that snippet, but it's probably a good idea to keep it with the rest of the snippet structure so it'll be easier for future use.

nastasha-solomon: Ok, can do. No objections to removing the comments -- I was wondering if we should keep them. Agree with moving the pull attribute up to 8.8.1. Even if there's no use for it at the moment, it's just good to keep it grouped for future use.

joepeeples (approved):
LGTM!

jmikell821 (approved):

Issue #3160, opened 2023-04-13
Exception API docs


DionAkkerman:
## Description 

1. Error in url in Docs for [exceptions-api-update-container](https://www.elastic.co/guide/en/security/current/exceptions-api-update-container.html) for the last couple versions.
Example url is "PUT <kibana host>:<port>/api/lists" should be "PUT api/exception_lists"

2. Filter description in [exceptions-api-find-exception-containers](https://www.elastic.co/guide/en/security/current/exceptions-api-find-exception-containers.html) could use improvement. The docs suggest using <field name>:<field value> but ignores that the fields are nested. To use "name" for example, <exception-list.attributes.name>:<field value> is required.

yctercero:
Please find comments below:

1. That's correct - should be `/api/exception_lists`

2. That is correct, that the filters should be formatted as described in the ticket description. However, this [PR](https://github.com/elastic/kibana/pull/138770) also added a `search` param which allows the user to more easily search using the simple query search outlined [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html#simple-query-string-syntax).

Here's a [test](https://github.com/elastic/kibana/blob/main/x-pack/test/lists_api_integration/security_and_spaces/tests/find_exception_list_items.ts#L60) that gives an example of using `search` and the expected results.

yctercero:
NOTE TO SELF: Expand on need for specifying SO type in addition to field name for `filter` and expand on how to use `search` param.

yctercero:
Updated text for the `filter` param:

>Filters the returned results according to the value of the specified field, using the <so type>.<field name>:<field value> syntax, where <field name> can be: 
>
>- name
>- type
>- created_by
>- updated_by
>
>And <so type> can be:
>- exception-list - Space aware exception lists. 
>- exception-list-agnostic - Global exception lists that are shared across spaces. 

Text for `search` param:

>Filters the returned results according to the [simple query string](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html#simple-query-string-syntax) filter.

PR #3456, opened 2023-06-11
[API docs] Expands on docs for finding exception containers

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/3160.

Previews:
- [Find exception containers](https://security-docs_3456.docs-preview.app.elstc.co/guide/en/security/master/exceptions-api-find-exception-containers.html): Expands on the docs for the `filter` param and adds the `search` param, which was added in `8.5`.


yctercero (approved):
👌🏽 Looks great, thanks!

docs/detections/api/exceptions/api-find-exception-containers.asciidoc
```diff
@@ -34,15 +34,22 @@ See {ref}/search-request-body.html#request-body-search-search-after[Search After
...
-specified field, using the `<field name>:<field value>` syntax, where
-`<field name>` can be:
+|`filter` |String a|Filters the returned results according to the value of the specified field. 
+
+Uses the `<so type.field name:field value>` syntax, where `<so type>` can be:
+
+* `exception-list`: Specify a space-aware exception list.
+* `exception-list-agnostic`: Specify an exception list that is shared across spaces.
+
+And `<field name>` can be:
```
joepeeples, line 44: Maybe omit the angle brackets -- it's a little unclear if they're literal syntax or functioning more like quotation marks, and sometimes they're used for the whole string and sometimes just for a part of it. The monospace format already makes the values stand out from normal text.

```suggestion
Uses the `so type.field name:field value` syntax, where `so type` can be:

* `exception-list`: Specify a space-aware exception list.
* `exception-list-agnostic`: Specify an exception list that is shared across spaces.

And `field name` can be:
```

joepeeples (commented):
Small formatting suggestion, otherwise LGTM!

PR #3503, opened 2023-06-28
Interactive investigation guides enhancements in 8.9 

joepeeples:
Resolves #3188.

Preview: [Launch Timeline from investigation guides](https://security-docs_3503.docs-preview.app.elstc.co/guide/en/security/master/interactive-investigation-guides.html)

Related: The callout about version compatibility needed to be revised for all versions going back to 8.6. But because this PR makes many other changes that don't need to be backported -- including relocating the file `investigation-guide-actions.asciidoc` -- updateing to 8.6-8.8 will need to be handled a little differently:
- This PR handles updates for `main` and `8.9` (the latter via Mergify).
- https://github.com/elastic/security-docs/pull/3505 will update `8.8`, then I'll use the backport tool to backport the change to `8.7` and `8.6`.

Issue #3188, opened 2023-04-26
Interactive investigation guides enhancements in 8.9


joepeeples:
Document the latest enhancements or changes to the interactive investigation guides feature (ongoing work done in 8.6 with https://github.com/elastic/security-docs/issues/2696 and in 8.7 with https://github.com/elastic/security-docs/issues/2955). 

Also, the interactive investigation guide feature is set to go GA in the future, and will probably have some licensing/subscription requirements. Exact release TBA.

Related:
- https://github.com/elastic/kibana/issues/153377
   From issue description:
    > Scope of this feature will now include 3 parts: 
    > 1. hardening of the rendered button, so that we optimistically use fields from old versions with sensible defaults, 
    > 2. ability to delete timeline notes so in the event that invalid markdown is saved as a note, it can be removed by users
    > 3. wrapping the markdown renderer in a React ErrorBoundary, to prevent any rendering exceptions and also display any relevant parsing errors to the user.

- Epic: https://github.com/elastic/security-team/issues/4824

## To Do

- [x] Add info on backwards compatibility, any other new developments
- [x] Update screenshot with new icon in editor
- [x] Remove "technical preview" badge from screenshots
- [x] Relocate page out of [Technical preview](https://www.elastic.co/guide/en/security/master/sec-experimental-intro.html) section

joepeeples:
Update: This issue originally included the "delete notes in Timeline" change (https://github.com/elastic/kibana/pull/154834), but I broke that out into a separate docs issue since the various features might not all release GA at the same time now: https://github.com/elastic/security-docs/issues/3224

joepeeples:
Pushed back to 8.9+

PR #3505, opened 2023-06-28
Backport updates to compatibility callout - investigation guides

joepeeples:
Backports the revised compatibility callout as part of #3188.

Doing this manually because the file `investigation-guide-actions.asciidoc` was moved in the main/8.9 PR (#3503), along with several other non-backportable changes that make it difficult to use Mergify or backport tool. Once this PR is merged to `8.8`, then I should be able to use the backport tool to update `8.7` and `8.6`. 🤞

[Preview](https://security-docs_3505.docs-preview.app.elstc.co/guide/en/security/8.8/interactive-investigation-guides.html)

### DIFF
--- a/docs/experimental-features/investigation-guide-actions.asciidoc
+++ b/docs/experimental-features/investigation-guide-actions.asciidoc
@@ -1,7 +1,7 @@
 [[interactive-investigation-guides]]
 == Interactive investigation guides
 
-IMPORTANT: This feature is available for {stack} versions 8.6.0 and newer. Additionally, interactive investigation guides are not compatible between release versions, so query buttons created in 8.6.0 won't render correctly in {version} and vice versa.
+IMPORTANT: Interactive investigation guides are compatible between {stack} versions 8.7.0 and later. Query buttons created in 8.6.x use different syntax and won't render correctly in later versions, and vice versa.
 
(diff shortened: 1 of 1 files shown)


benironside (approved):
Looks good to me, but as with the related PR I think the IMPORTANT callout would make more sense after the "Detection rule investigation guides..." paragraph. On the other hand I guess it depends why users are coming to this part of docs. I hope that it's not mostly to figure out why their rules from 8.6 are broken, but I can't say for sure.

joepeeples:
> Looks good to me, but as with the related PR I think the IMPORTANT callout would make more sense after the "Detection rule investigation guides..." paragraph. On the other hand I guess it depends why users are coming to this part of docs. I hope that it's not mostly to figure out why their rules from 8.6 are broken, but I can't say for sure.

I agree, it's worth reordering the callout, especially since we're already touching the file in these versions anyway. The [original callout in 8.6](https://www.elastic.co/guide/en/security/8.6/interactive-investigation-guides.html) was a lot simpler, and followed the example of [host/user risk score](https://www.elastic.co/guide/en/security/8.6/host-risk-score.html). But as we've needed to add more info to the callout, it does flow much better _below_ the main intro, not above it.

joepeeples:
https://github.com/Mergifyio backport 8.7 8.6

PR #3541, opened 2023-07-07
[DOCS] Add category field to cases

lcawl:
Relates to https://github.com/elastic/kibana/issues/153835

This PR updates https://www.elastic.co/guide/en/security/current/cases-open-manage.html so that the optional category field is mentioned alongside the similar tags field. Unlike tags, you can choose only a single category value for each case, which is why it's not pluralized in some of these phrases.

This PR also updates three screenshots that are now automatically generated per https://github.com/elastic/kibana/pull/161412 and https://github.com/elastic/kibana/pull/161485


nastasha-solomon (approved):
Looks great! The only remaining screenshot that needs to be updated is the one after step 2 [here](https://www.elastic.co/guide/en/security/8.8/signals-to-cases.html#signals-to-existing-cases) (`add-alert-to-existing-case.png`).

Thanks as always for updating these! Thanks as well for flagging auto-generated screenshots. The comments make it _much_ easier to track/know which images are auto-crated.

lcawl:
> Looks great! The only remaining screenshot that needs to be updated is the one after step 2 [here](https://www.elastic.co/guide/en/security/8.8/signals-to-cases.html#signals-to-existing-cases) (`add-alert-to-existing-case.png`).

Good catch!  I've refreshed those screenshots in 399000506dde5b38a6a3cfa3066e79b78c0cadb7
I also added frontmatter for the docs migration, but feel free to change those values in future updates.

Issue #3276, opened 2023-05-16
[Docs][Exceptions] - Auto fill exceptions from alert data


yctercero:
## Description 

This issue aims to enhance the feature of auto-populating Rule Exceptions that are created from Alerts.

Currently, Rule Exceptions can be created from both the Alerts table and the Alert summary flyout action button. The goal is to automatically populate the Rule Exceptions with the relevant fields extracted from the Alert. However, certain fields that are deemed irrelevant to the Rule Exception need to be stripped out as described [here](https://github.com/elastic/security-team/issues/6405).

**Version**: This feature should be in 8.9 
**PR**: https://github.com/elastic/kibana/pull/159029  

### Doc updates 

- [Add exceptions to a rule](https://www.elastic.co/guide/en/security/master/add-exceptions.html#detection-rule-exceptions): Update steps 1 and 2 (screenshot in step 2 included)
- [Add Elastic endpoint exceptions](https://www.elastic.co/guide/en/security/master/add-exceptions.html#endpoint-rule-exceptions): Screenshot in step 1 and possibly text in step 2.

## Notes
- Will also need to update the endpoint exception [docs](https://www.elastic.co/guide/en/security/current/add-exceptions.html#endpoint-rule-exceptions) to show that conditions (fields and values) for endpoint exceptions are also auto-populated when users create a new exception from the **More Actions** menu or Alerts table. 
   - The fix (https://github.com/elastic/kibana/pull/159908) for this was only backported to 8.9/8.10 so will likely only need to update the 8.9 docs.  
- Condition fields and values are only auto-populated when users create a new exception from an alert (**Take action > Add rule exception**) or from the Alert table (**More Actions menu (…​) > select Add rule exception**).
   - https://github.com/elastic/kibana/pull/159908 adds functionality to the Take action menu. 
   - Condition fields/values _are not_ auto-populated when users add an exception to a rule from the Shared Exception Lists page.

PR #3544, opened 2023-07-07
Auto fill exceptions from alert data

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/3276.

Preview:

- [Add exceptions to a rule](https://security-docs_3544.docs-preview.app.elstc.co/guide/en/security/master/add-exceptions.html#detection-rule-exceptions): Broke step two into two parts and added a note after step 2b that explained when exceptions are auto-populated.


WafaaNasr (approved):
LGTM!!

docs/detections/add-exceptions.asciidoc
```diff
@@ -53,7 +53,12 @@ image::images/rule-exception-tab.png[Detail of rule exceptions tab]
 .. Click *Create shared exception list* -> *Create exception item*. 
 --
 
-. In the *Add rule exception* flyout, name the exception and add conditions that define the exception. When the exception's query conditions are met (the query evaluates to `true`), rules do not generate alerts even when other rule criteria are met.
+. In the *Add rule exception* flyout, do the following: 
+
+.. Name the exception. 
+.. Add conditions that define the exception. When the exception's query conditions are met (the query evaluates to `true`), rules do not generate alerts even when other rule criteria are met.
++
+NOTE: Exception conditions are auto-populated with relevant alert data when you create a new exception from the alert's details flyout or from the Alert table.
```
WafaaNasr, line 61: Great description Nastasha!.
It is indeed accurate that there is no need to mention the "Highlighted fields" since we only utilize the relevant data from them only.

nastasha-solomon: Note to self that I might need to expand this note to include the comment that's also auto-generated when users create a new exception from the Alerts table or flyout.

nastasha-solomon:
```suggestion
NOTE: When you create a new exception from the alert's details flyout or from the Alert table, exception conditions are auto-populated with relevant alert data. A comment describing this action is also automatically added to the **Add comments** section.
```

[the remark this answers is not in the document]
nastasha-solomon: Thank you! I added your feedback in [059a805](https://github.com/elastic/security-docs/pull/3544/commits/059a805dedf498622f1f7bb9fdb15b43a51c765c).

natasha-moore-elastic (approved):

docs/detections/add-exceptions.asciidoc
```diff
@@ -53,7 +53,12 @@ image::images/rule-exception-tab.png[Detail of rule exceptions tab]
 .. Click *Create shared exception list* -> *Create exception item*. 
 --
 
-. In the *Add rule exception* flyout, name the exception and add conditions that define the exception. When the exception's query conditions are met (the query evaluates to `true`), rules do not generate alerts even when other rule criteria are met.
+. In the *Add rule exception* flyout, do the following: 
+
+.. Name the exception. 
+.. Add conditions that define the exception. When the exception's query conditions are met (the query evaluates to `true`), rules do not generate alerts even when other rule criteria are met.
```
joepeeples, line 59: I had a little difficulty following the second sentence -- there's a lot of passive voice, maybe could be streamlined. I also thought "other rule criteria" sounded a little vague, like maybe some rule criteria are involved while other criteria might not be, but I think it's just the rule itself.
```suggestion
.. Add conditions that define the exception. When the exception's query evaluates to `true`, rules don't generate alerts even when their criteria are met.
```

docs/detections/add-exceptions.asciidoc
```diff
@@ -53,7 +53,12 @@ image::images/rule-exception-tab.png[Detail of rule exceptions tab]
...
++
+NOTE: When you create a new exception from the alert's details flyout or from the Alert table, exception conditions are auto-populated with relevant alert data. A comment describing this action is also automatically added to the **Add comments** section.
```
joepeeples, line 61: We could probably simplify this -- I get that naming each particular context adds specificity, but it gets the sentence off to a long, slow start, and users might not need to know every context (X and Y parts of the UI), just the general context (alerts).
```suggestion
NOTE: When you create a new exception from an alert, exception conditions are auto-populated with relevant alert data. A comment describing this action is also automatically added to the **Add comments** section.
```

joepeeples (commented):
Added a couple streamlining edits to consider, thanks! 

Also, I noticed on the preview that step 2 is super long, with multiple commands that say basically the same thing:

"Add conditions that define the exception."
"Add conditions that define when the exception prevents alerts."

Maybe the second one should be a separate step? Or revise so they're not so similar?

image

docs/detections/add-exceptions.asciidoc
nastasha-solomon: Yeah good point, especially when the necessary context is provided in the previous step. I'll commit this edit - thanks!

Issue #3497, opened 2023-06-27
Negative correlation in event correlation (EQL) rules


joepeeples:
## Description 

Event correlation (EQL) rules now support negative correlation for missing events. When writing an EQL query for an event correlation rule, users can specify events that _should not_ happen in a sequence. 

## Related
* Security issue: https://github.com/elastic/security-team/issues/2835
* Added to Elasticsearch: https://github.com/elastic/elasticsearch/pull/92348
   - Note: [maxspan] is required for sequences with missing events queries (per [here](https://github.com/elastic/elasticsearch/pull/92348/files#diff-123105fcd0030e62f4100ab1424e1e978bcffa93eb64169867557eb3136c0fe1R338))
* Docs issue for Elasticsearch: https://github.com/elastic/platform-docs-team/issues/72

## Areas to update

Technically this is just a change to the EQL language, but it's requested/popular enough in Security to warrant an update on our side.

- [x] Release notes: Will be documenting in the [8.9.1 release notes](https://github.com/elastic/security-docs/pull/3690).
- [ ] What's new: @jmikell821 to file a doc issue or PR for these changes. 
- [x] Create rule instructions - Mention negative correlation, ~add example ([suggested example](https://github.com/elastic/eql/issues/21#issuecomment-1599745058))~.
- [x] Elasticsearch docs - Will be updated by Platform docs team, but might want to coordinate for consistency, links. Update: EQL docs published.

nastasha-solomon:
Feature has been moved out of 8.9. Might be implemented in 8.10, hence the version label update.

cc: @yctercero

nastasha-solomon:
Re-opening since release notes haven't been merged yet.

nastasha-solomon:
Merged doc updates and 8.9.1 release notes.

PR #3709, opened 2023-08-10
Negative correlation in event correlation (EQL) rules

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/3497.

[Preview](https://security-docs_3709.docs-preview.app.elstc.co/guide/en/security/master/rules-ui-create.html#create-eql-rule) - Made the following changes:
- **Step 1a**: Specified that the rule queries indices and data views for events (not alerts)
- **Step 1b**: Slightly revised the language and added a tip about using the syntax for negative correlation


paulewing (approved):
LGTM!

docs/detections/rules-ui-create.asciidoc
```diff
@@ -206,8 +206,10 @@ IMPORTANT: Alerts created by threshold rules are synthetic alerts that do not re
 [[create-eql-rule]]
 ==== Create an event correlation rule
 . To create an event correlation rule using EQL, select *Event Correlation*, then:
-.. Define which {es} indices or data view the rule searches for alerts.
-.. Add an {ref}/eql-syntax.html[EQL statement] used to detect alerts.
+.. Define which {es} indices or data view the rule searches when querying for events.
+.. Write an {ref}/eql-syntax.html[EQL query] that searches for matching events or a series of matching events.
++
+TIP: When searching for a series of matching events, use the {ref}/eql-syntax.html#eql-missing-events[missing events] syntax to specify events that should _not_ happen.
```
joepeeples, line 212: Maybe a different word at the end? It's unclear what it means if an event "should not happen." Does the syntax exclude events from the search?
```suggestion
TIP: When searching for a series of matching events, use the {ref}/eql-syntax.html#eql-missing-events[missing events] syntax to specify events that should _not_ be matched.  
```

nastasha-solomon: @joepeeples if I'm understanding the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/master/eql-syntax.html#eql-missing-events) correctly, negative correlation syntax should return events that don't match the conditions you specify. So it'd be like saying, "Find me all the events that _don't match_ this set of criteria." @paulewing @yctercero am I understanding this correctly? 

If I am, how does the following look:

```suggestion
TIP: To find events that don't match your specified conditions, use the {ref}/eql-syntax.html#eql-missing-events[missing events] syntax.  
```

jmikell821: @nastasha-solomon I like this second option better. 👍

nastasha-solomon: Revising this based on @paulewing's feedback on Slack:

 ```suggestion
TIP: To find events that are missing in a sequence, use the {ref}/eql-syntax.html#eql-missing-events[missing events] syntax.  
```

yctercero (approved):

paulewing (approved):

PR #3710, opened 2023-08-10
Remove book-scoped attributes in AsciiDoc [Serverless]

### LINKED ISSUE #3670: Remove book-scoped attributes in AsciiDoc [Serverless]
## Description 

We need to remove all instances of [book-scoped attributes](https://docs.elastic.dev/migration/book-level/attributes) throughout `security-docs` to ensure smooth migration to MDX. While our files can still be migrated to MDX with book-scoped attributes, they create a dependency in the migrated files that can cause additional problems.

We can replace these with either the equivalent global [shared attributes](https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc) or hard-written values.

This affects migration of docs for Serverless, but will also affect general migration to MDX for non-Serverless docs. We'll need to backport to `8.9`, which has already been migrated to MDX but might receive updates, and we'll want those updates to be free of these attributes. 

## Acceptance Test Criteria

Book-scoped attributes are defined in the book's main index file: [security-docs/docs/index.asciidoc](https://github.com/elastic/security-docs/blob/89aecc141d6374836b8dd3ee43ed314afb6ecc62/docs/index.asciidoc?plain=1#L7-L17). 

```
:es-sec:       {elastic-sec}
:es-sec-app:   {security-app}
:es-sec-ui:    {elastic-sec} UI
:es-sec-endpoint: {elastic-defend}
:siem-soln:    {elastic-sec}
:siem-app:     {security-app}
:siem-ui:      {es-sec-ui}
:ml-dir:       {stack-docs-root}/docs/en/stack/ml
:beats-dir:    {beats-root}
:kibana-dir:   {kibana-root}/docs
```

### To do

- In general, follow the [migration guide's process](https://docs.elastic.dev/migration/book-level/attributes#pre-migration-tasks) for removing book-scoped attributes.
- Confirm if any other files (indexes or otherwise) contain additional attribute definitions, and add them to your to-do list.
   - Update: Checked the repo, found no other additional attributes defined in other files.
- Search across the repo for instances where each attribute is used in content, and replace with a global/shared attribute. For example, search for `{es-sec}`, and replace it with `{elastic-sec}`.
- When backporting to `8.9`, do _an additional search_ to make sure there aren't any instances that exist there, but later got revised out of `main`/`8.10`. We want to be double-sure that we have zero book-scoped attributes.

joepeeples:
Resolves #3670.

Removes the dependency on book-scoped attributes from the Elastic Security docs, per [migration guide requirement](https://docs.elastic.dev/migration/book-level/attributes#pre-migration-tasks). In most cases, the attribute is replaced with an equivalent site-scoped [shared attribute](https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc), or a hard-written value.

We'll need to backport to `8.9`, which has already been migrated to MDX but might receive updates, and we'll want those updates to be free of these attributes.

REVIEWERS: Please try pulling down this branch and building locally, to make sure nothing breaks. And probably no need to preview every single affected page; just a few spot-checks should cover it.


docs/detections/prebuilt-rules-management.asciidoc
```diff
@@ -6,7 +6,7 @@
 :frontmatter-tags-content-type: [how-to]
 :frontmatter-tags-user-goals: [manage]
 
-Follow these guidelines to start using the {es-sec-app}'s <<prebuilt-rules, prebuilt rules>>, keep them updated, and make sure they have the data needed to run successfully. 
+Follow these guidelines to start using the {security-app}'s <<prebuilt-rules, prebuilt rules>>, keep them updated, and make sure they have the data needed to run successfully.
```
nastasha-solomon, line 9: For my own edification, when should we use {security-app} over {elastic-sec}? I feel like {elastic-sec} would work here too, and might even be more appropriate to use since the security rules are actually from Elastic.

docs/detections/prebuilt-rules/tune-rule-signals.asciidoc
```diff
@@ -1,7 +1,7 @@
 [[tuning-detection-signals]]
 == Tune detection rules
 
-Using the {es-sec-app}, you can tune prebuilt and custom detection rules to optimize alert generation. To reduce noise, you can:
+Using the {security-app}, you can tune prebuilt and custom detection rules to optimize alert generation. To reduce noise, you can:
```
nastasha-solomon, line 4: Same question as above - why is {security-app} more appropriate than {elastic-sec} in this situation?

nastasha-solomon (approved):
Overall, looks good! I just have the one question for you below - thanks!

docs/detections/prebuilt-rules-management.asciidoc
joepeeples: @nastasha-solomon Good question, I added to the team sync agenda, since it should probably be a separate discussion with the whole team. This PR is just removing the dependency on book-scoped attributes. But I think at one point, Elastic might have wanted to call the solutions "apps", though I'm not sure that's still the preferred term.

benironside (approved):
LGTM!

PR #3712, opened 2023-08-11
Update issue templates for known issues. 

jmikell821:


nastasha-solomon (approved):

.github/ISSUE_TEMPLATE/known-issue.md
```diff
@@ -0,0 +1,28 @@
...
+about: 'Report a known issue for us to include in release notes. '
+title: "[Known Issue] "
+labels: known-issue, release-notes
+assignees: ''
+
+---
+
+## Description
+
+Please describe what the known issue is.
```
joepeeples, line 12: Maybe ask for a little more info to help us prioritize?
```suggestion
Please describe what the known issue is, and the severity of its impact to customers.
```

joepeeples (approved):
One suggestion for more info, LGTM!

nastasha-solomon:
@jmikell821 sorry - just a few additional questions and thoughts from my end.

It might be helpful to outline guidelines for deciding whether a bug should be doc'd as a known issue. I think in general, devs understand that users need to be informed if there’s a bug in the UI or a bug in a feature that creates unexpected behavior. What seems to be less clear is whether they should also consider other factors like:
- The likelihood of a user encountering the bug (some bugs are discovered when QA or devs are testing edge cases)
- The scope and criticality of the bug (some bugs are pretty small in scope and aren't blocking users from interacting with features)

Here's my perspective:
- If the bug is a security vulnerability, causes data loss, or is severe enough that it stops vital functionality from working, document it as a known issue.
- If the bug is a niche edge-case that users haven't reported yet and will be quickly fixed, either put it in a knowledge article or the internal troubleshooting docs that devs maintain.
- If the bug won't be fixed, note the limitation in the feature docs. 

So my questions for you (and really the entire @elastic/security-docs team) are: WDYT about adding some guidelines for determining whether a bug should be doc'd as a known issue? If we do add guidelines, what should they be?

jmikell821:
@nastasha-solomon this are all great points! Let me draft something in Google and we can attach it to the know issue template here. 

> @jmikell821 sorry - just a few additional questions and thoughts from my end.
> 
> It might be helpful to outline guidelines for deciding whether a bug should be doc'd as a known issue. I think in general, devs understand that users need to be informed if there’s a bug in the UI or a bug in a feature that creates unexpected behavior. What seems to be less clear is whether they should also consider other factors like:
> 
> * The likelihood of a user encountering the bug (some bugs are discovered when QA or devs are testing edge cases)
> * The scope and criticality of the bug (some bugs are pretty small in scope and aren't blocking users from interacting with features)
> 
> If I were to suggest some guidelines, here's what they'd be:
> 
> * If the bug is a security vulnerability, causes data loss, or is severe enough that it stops vital functionality from working, document it as a known issue.
> * If the bug is a niche edge-case that users haven't reported yet and will be quickly fixed, either put it in a knowledge article or the internal troubleshooting docs that devs maintain.
> * If the bug won't be fixed, note the limitation in the feature docs.
> 
> So my questions for you (and really the entire @elastic/security-docs team) are: WDYT about adding some guidelines for determining whether a bug should be doc'd as a known issue? If we do add guidelines, what should they be?

Issue #3572, opened 2023-07-17
Add a note in 8.8 docs about known legacy action migration issue


e40pud:
## Description

In 8.8 we introduced conditional actions and moved legacy actions migration code to the alerting framework.

As part of these changes we introduced [the issue](https://github.com/elastic/kibana/issues/157462) which can happen during the migration. The issue happens when the frequency of legacy action is shorter than the rule execution interval. In this case during the migration alerting framework will convert and delete legacy actions and then do validation (action frequency cannot be shorter than rule interval) and throw an exception which leads to data loss (all actions will be lost in this case).

We will add [the fix](https://github.com/elastic/kibana/pull/160798) for this issue in `v8.9.1` and need to report a known issue for `v8.8`.

### There are few possible workarounds:
1. Users can manually set rule interval to be greater or equal to rule's interval and save the rule.
2. Users can export a rule, fix it manually and import it again.
3. Users can remove legacy action from rule in UI. Save rule, and try to create new action afterwards for this rule.
4. Users can just re-create the whole rule from scratch.

nastasha-solomon:
Am using the summary that was reviewed/confirmed [here](https://github.com/elastic/security-docs/pull/3536#discussion_r1271544745).

PR #3722, opened 2023-08-14
Add a note in 8.8 docs about known legacy action migration issue

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/3572.

[Preview](https://security-docs_3722.docs-preview.app.elstc.co/guide/en/security/master/release-notes-header-8.8.0.html) - Added the same known issue summary to all 8.8.x releases and the 8.9.1 release notes.


e40pud (approved):
Looks great! Thanks for adding docs for this issue!

jmikell821 (approved):
LGTM!

e40pud:
@nastasha-solomon I just realised that the documentation we are trying to add here are for the fix that we added in 8.9.1+, in 8.8 we have a data loss issue which we should report as a known issue.

In 8.8 `Rule changes can’t be saved if the rule’s action frequency is shorter than the rule’s run interval.`, but also such legacy actions will be removed (causing data loss).

docs/release-notes/8.8.asciidoc
```diff
@@ -9,6 +9,7 @@
 [[known-issue-8.8.2]]
 ==== Known issues
 
+* Rule changes can't be saved if the rule's action frequency is shorter than the rule's run interval.
```
nastasha-solomon, line 12: @e40pud thanks for double-checking this. How does look?

```suggestion
* Rule changes can't be saved and existing rule actions are removed if the action's frequency is shorter than the rule's run interval.
```

e40pud: Perfect! Thank you!

PR #3772, opened 2023-08-21
Fix several typos in "Configure an integration policy for Elastic Defend"

gabriellandau:
- [x] Change several Agent references to Defend/Endpoint.  Defend/Endpoint provides these features, not Agent itself.
- [x] Correct a typo related to malware detect/prevent quarantine.


joepeeples (approved):
One additional fix, otherwise LGTM! Thanks for this update @gabriellandau!

joepeeples:
FYI, I added the `v8.9.0` label since that's what the Mergify backporting service looks for.

PR #3818, opened 2023-08-28
Revise cross-cluster search docs per ES changes

### LINKED ISSUE #3787: Revise cross-cluster search docs per ES changes
## Description 

Cross-cluster search is changing in Elasticsearch as of 8.10, so we'll need to update our [Cross-cluster search and detection rules](https://www.elastic.co/guide/en/security/master/rules-cross-cluster-search.html) page accordingly. This issue follows up on https://github.com/elastic/security-docs/pull/3742, which fixes some broken links to the new ES docs and suggests further revisions:

> 8.10 introduces a new mechanism to configure remote clusters (API key based), which will live next to the existing mechanism (certificate based). As a result, the ES remote cluster docs will be restructured. 
> 
> [ . . . ] 
> 
> By the way, I think the "Set up cross-cluster search in detection rules" section needs to be rewritten to accommodate the new mechanism. This PR only changes the links though, to unblock https://github.com/elastic/elasticsearch/pull/98330.
> 

## Notes

- Follow up with @abdonpijpelink on what needs to be changed.
- Include Security engineering to verify changes to the detection rule process.

joepeeples:
Resolves #3787.

Preview: [Cross-cluster search and detection rules](https://security-docs_3818.docs-preview.app.elstc.co/guide/en/security/master/rules-cross-cluster-search.html)


docs/detections/rules-cross-cluster-search.asciidoc
```diff
@@ -9,9 +9,11 @@
 
 This section explains the general process for setting up cross-cluster search in detection rules. For specific instructions on each part of the process, refer to the linked documentation.
 
-. On the local cluster, {ref}/remote-clusters.html[set up a connection] to the remote cluster. Note the unique name that you give to the remote cluster, because you'll need to use it throughout this process.
+NOTE: This procedure uses TLS certificate authentication to add remote clusters. {stack} 8.10.0 introduces an alternate method using {ref}/remote-clusters-api-key.html[API key authentication], but it is not yet included in the following steps.
 
-. On both the local and remote clusters, {ref}/remote-clusters.html[create a
+. On the local cluster, {ref}/remote-clusters-cert.html#remote-clusters-connect-cert[set up a connection] to the remote cluster. Note the unique name that you give to the remote cluster, because you'll need to use it throughout this process.
```
abdonpijpelink, line 14:
```suggestion
. On the local cluster, {ref}/remote-clusters-cert.html[set up a connection] to the remote cluster. Note the unique name that you give to the remote cluster, because you'll need to use it throughout this process.
```

abdonpijpelink: I would link to the top-level page for setting up remote clusters with certificates, otherwise users may skip the setting up trust part.

abdonpijpelink (approved):
LGTM. I've left a suggestion.

docs/detections/rules-cross-cluster-search.asciidoc
```diff
@@ -9,9 +9,11 @@
 
 This section explains the general process for setting up cross-cluster search in detection rules. For specific instructions on each part of the process, refer to the linked documentation.
 
-. On the local cluster, {ref}/remote-clusters.html[set up a connection] to the remote cluster. Note the unique name that you give to the remote cluster, because you'll need to use it throughout this process.
+NOTE: This procedure uses TLS certificate authentication to add remote clusters. {stack} 8.10.0 introduces an alternate method using {ref}/remote-clusters-api-key.html[API key authentication], but it is not yet included in the following steps.
```
vitaliidm, line 12: by mentioning 
>  but it is not yet included in the following steps.
do we imply this method is not supported/verified yet for cross-cluster rule runs?

FYI: I have raised support of it on out team meeting and we going to take action to find out who would be the best fir to verify rules cross-cluster behaviour

joepeeples: > by mentioning
> 
> > but it is not yet included in the following steps.
> 
> do we imply this method is not supported/verified yet for cross-cluster rule runs?

@vitaliidm Yes, I wouldn't want to document anything more detailed until the team has verified this new method. Maybe this needs to be more explicit, something like "but it is not yet supported for detection rules"? I would update the statement, of course, once the team verifies the process.

docs/detections/rules-cross-cluster-search.asciidoc
joepeeples: @abdonpijpelink Good point, thanks. I was trying to avoid pushing the reader to another long page with another set of steps for them to follow, but it sounds like it's somewhat unavoidable. Maybe I can give some extra context to help the reader target what they really need on that other page:
```suggestion
. On the local cluster, {ref}/remote-clusters-cert.html[establish trust and set up a connection] to the remote cluster. Note the unique name that you give to the remote cluster, because you'll need to use it throughout this process.
```

docs/detections/rules-cross-cluster-search.asciidoc
joepeeples:
```suggestion
NOTE: This procedure uses TLS certificate authentication to add remote clusters. {stack} 8.10.0 introduces an alternate method using {ref}/remote-clusters-api-key.html[API key authentication], but it is not yet supported for detection rules.
```

vitaliidm: I think it is better and place us on a safe side, until it verified.
Also, according to https://www.elastic.co/guide/en/elasticsearch/reference/8.10/remote-clusters-api-key.html it is in `beta` stage. So we might rephrase, we strongly recommend the old way to set-up clusters, and cannot guarantee the new on would work. But either way is fine by me

vitaliidm (approved):

PR #3914, opened 2023-09-11
Add new terms rule type to Detections API create/update rule docs

marshallmain:
Adds API docs for the new terms rule type, introduced in https://github.com/elastic/kibana/pull/134526 (stack v8.4.0). Support for multiple values in `new_terms_fields` was added in 8.6 (https://github.com/elastic/kibana/pull/143943), so if we backport as far as 8.4 and 8.5 we should modify the `Must contain 1-3 field names.` text to reflect the limitation (something like `Must contain 1 field name.` instead).


natasha-moore-elastic (commented):
Thanks for raising the PR, @marshallmain! LGTM, just a couple of suggestions:

1. Can we add the following text to the list of rule types at the start of the `rules-api-create` page (on line 36, after Machine learning rules):
 
> **New terms**: Generates an alert for each new term detected in source documents within a specified time range.

2. Can we add request and response examples for creating a new terms rule to the `rules-api-create` page? Just to keep the consistency, as the page contains examples for all other rule types.

docs/detections/api/rules/rules-api-update.asciidoc
```diff
@@ -166,6 +167,20 @@ uses both `and` and `or` logic.
...
+
+[width="100%",options="header"]
+|==============================================
+|Name |Type |Description
+
+|new_terms_fields |String[] |Fields to monitor for new values. Must contain 1-3 field names.
+
+|history_window_start |String |Start date to use when checking if a term has been seen before.
+Supports relative dates, e.g. `now-30d` will search the last 30 days of data when checking if a term
+is new.
```
natasha-moore-elastic, line 180: [Same comment as for `rules-api-update`](https://github.com/elastic/security-docs/pull/3914/files#r1322987743).

docs/detections/api/rules/rules-api-create.asciidoc
```diff
@@ -250,6 +252,21 @@ uses both `and` and `or` logic.
...
+
+[width="100%",options="header"]
+|==============================================
+|Name |Type |Description
+
+|new_terms_fields |String[] |Fields to monitor for new values. Must contain 1-3 field names.
+
+|history_window_start |String |Start date to use when checking if a term has been seen before.
+Supports relative dates, e.g. `now-30d` will search the last 30 days of data when checking if a term
+is new.
```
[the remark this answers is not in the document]
marshallmain: good question, it technically supports absolute dates as well - and we use that for testing - but it's not something I would want to recommend even to API power users. Do we have a way to communicate "this is possible but use at your own risk"? The risk with absolute dates being used for the history window is that the history window then grows indefinitely as time goes on so rule performance is likely to gradually degrade as it has to query more and more data. If some of that history data moves to cold/frozen tier then performance will take a huge hit at that point as well.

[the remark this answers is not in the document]
natasha-moore-elastic: I see, thanks for that context! In that case, we can probably omit mentioning support for absolute dates, and include a line that states we don't recommend using them. For example:

> Supports relative dates – for example, `now-30d` will search the last 30 days of data when checking if a term
> is new. We do not recommend using absolute dates, which can cause issues with rule performance due to increased data querying over time.

docs/detections/api/rules/rules-api-create.asciidoc
```diff
@@ -250,6 +252,22 @@ uses both `and` and `or` logic.
...
+===== Required fields for new terms rules
+
+[width="100%",options="header"]
+|==============================================
+|Name |Type |Description
+
+|new_terms_fields |String[] |Fields to monitor for new values. Must contain 1–3 field names.
+
+|history_window_start |String |Start date to use when checking if a term has been seen before.
+Supports relative dates – for example, now-30d will search the last 30 days of data when checking if a term
```
natasha-moore-elastic, line 265:
```suggestion
Supports relative dates – for example, `now-30d` will search the last 30 days of data when checking if a term
```

docs/detections/api/rules/rules-api-update.asciidoc
```diff
@@ -166,6 +167,21 @@ uses both `and` and `or` logic.
...
+===== Fields required for new terms rules `PUT` calls
+
+[width="100%",options="header"]
+|==============================================
+|Name |Type |Description
+
+|new_terms_fields |String[] |Fields to monitor for new values. Must contain 1–3 field names.
+
+|history_window_start |String |Start date to use when checking if a term has been seen before.
+Supports relative dates – for example, now-30d will search the last 30 days of data when checking if a term
```
natasha-moore-elastic, line 179:
```suggestion
Supports relative dates – for example, `now-30d`will search the last 30 days of data when checking if a term
```

natasha-moore-elastic:
> Thanks for raising the PR, @marshallmain! LGTM, just a couple of suggestions:
> 
> 1. Can we add the following text to the list of rule types at the start of the `rules-api-create` page (on line 36, after Machine learning rules):
> 
> > **New terms**: Generates an alert for each new term detected in source documents within a specified time range.
> 
> 2. Can we add request and response examples for creating a new terms rule to the `rules-api-create` page? Just to keep the consistency, as the page contains examples for all other rule types.

Hey @marshallmain, just flagging these two points again in case you missed them.

docs/detections/api/rules/rules-api-create.asciidoc
```diff
@@ -1241,3 +1283,48 @@ Example response for an indicator match rule:
...
+  "risk_score": 50,
+  "risk_score_mapping": [],
+  "name": "New User IP Detected",
+  "references": [],
+  "severity": "medium",
+  "severity_mapping": [],
+  "updated_by": "elastic",
+  "tags": [],
+  "to": "now",
+  "type": "threat_match",
```
natasha-moore-elastic, line 1314: if `type` is `new_terms` in the request, should it have the same value here?

docs/detections/api/rules/rules-api-create.asciidoc
```diff
@@ -1241,3 +1283,48 @@ Example response for an indicator match rule:
...
+  "description": "Detects a user associated with a new IP address",
+  "enabled": true,
+  "false_positives": [],
+  "from": "now-6m",
+  "id": "eb7225c0-566b-11ee-8b4f-bbf3afdeb9f4",
+  "immutable": false,
+  "interval": "5m",
+  "rule_id": "c6f5d0bc-7be9-47d4-b2f3-073d22641e30",
+  "max_signals": 100,
+  "risk_score": 50,
```
natasha-moore-elastic, line 1305: if `risk_score` is `21` in the request, should it have the same value here?

docs/detections/api/rules/rules-api-create.asciidoc
```diff
@@ -923,6 +942,29 @@ logic. In this example, both the event's `destination.ip` and
...
+unusual network connections:
+
+[source,console]
+--------------------------------------------------
+POST api/detection_engine/rules
+{
+  "risk_score": 21,
+  "description": "Detects a user associated with a new IP address",
+  "name": "New User IP Detected",
+  "severity": "medium",
```
natasha-moore-elastic, line 957: Should `severity` correspond to the `risk_score` value, i.e. should it be `low` as `risk_score` is `21` ?

docs/detections/api/rules/rules-api-create.asciidoc
```diff
@@ -923,6 +942,29 @@ logic. In this example, both the event's `destination.ip` and
 <4> Sibling `entries` are evaluated using `or` logic. An alert is generated when
 at least one `entries` object evaluates to `true`.
 
+*Example 6*
+
+New terms rule that creates alerts when the Windows `rundll32.exe` process makes
+unusual network connections:
```
natasha-moore-elastic, line 948: Just checking if this is correct as it doesn't seem to correspond to what the example is showing?

natasha-moore-elastic (approved):

PR #4037, opened 2023-10-13
[DOCS] Refresh case settings screenshots

lcawl:
Relates to https://github.com/elastic/kibana/issues/160236

This PR refreshes the screenshots and text in the Security > Cases documentation to reflect the change of the label from "Edit external connection" to "Settings".
It also makes use of a newly automated screenshot, which is generated by https://github.com/elastic/kibana/pull/168774

There is another screenshot that would need to be refreshed if it's kept, but for now I've commented it out.


nastasha-solomon (approved):
LGTM - thanks! 🚀🚀🚀

Issue #3817, opened 2023-08-28
[Detection Engine][ESQL] - Document new rule type


yctercero:
## Description 

Ticket for docs updates that will be required for new ESQL rule type.

Related:
- https://github.com/elastic/security-team/issues/7215
- https://github.com/elastic/kibana/pull/165450

## Copy work

- https://github.com/elastic/ux-writing/issues/41
- https://github.com/elastic/kibana/pull/168611

## Notes
To keep in mind:
- UI changes (rule creation/edit/details)
- API changes
- Will be tech preview

## Doc updates

- [About detection rules | Rule types](https://www.elastic.co/guide/en/security/master/about-rules.html#rule-types): 
   - [x] Add a bullet to the end of the list for the new ES|QL rule. 
- [Create a detection rule](https://www.elastic.co/guide/en/security/master/rules-ui-create.html): 
   - [x] Add a section for creating an ESQL rule. \
   - [x] Explain the two types of queries that users can write: aggregating and non-aggregrating  
- [Create rule API docs](https://www.elastic.co/guide/en/security/master/rules-api-create.html): 
   - [x] @vitaliidm making doc updates via https://github.com/elastic/security-docs/pull/4052

nastasha-solomon:
Temporarily re-opening since the PR still needs an editorial review and PM review.

PR #4062, opened 2023-10-17
ES|QL rule docs

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/3817.

Previews:
- [About detection rules | Rule types](https://security-docs_4062.docs-preview.app.elstc.co/guide/en/security/master/about-rules.html#rule-types) - Added a new bullet for ES|QL rules.
- [Create a detection rule | Create an ES|QL rule](https://security-docs_4062.docs-preview.app.elstc.co/guide/en/security/master/rules-ui-create.html#create-esql-rule) - Provides steps for creating an ES|QL rule, breaks down the types of queries users can make, and explains how to turn on alert deduping for non-aggregating ESQL rules.


docs/detections/about-rules.asciidoc
```diff
@@ -42,6 +42,9 @@ TIP: You can also use value lists as the indicator match index. See <<indicator-
 
 * <<create-new-terms-rule, *New terms*>>: Generates an alert for each new term detected in source documents within a specified time range. You can also detect a combination of up to three new terms (for example, a `host.ip` and `host.id` that have never been observed together before).
 
+* <<create-esql-rule, *ES|QL terms*>>: ES|QL-based rule, which searches the defined indices and
```
nastasha-solomon, line 45:
```suggestion
* <<create-esql-rule, *ES|QL terms*>>: Uses the Elasticsearch Query Language (ES|QL) to find events in the defined data sources and aggregate search results. 
```

docs/detections/about-rules.asciidoc
```diff
@@ -42,6 +42,12 @@ TIP: You can also use value lists as the indicator match index. See <<indicator-
 
 * <<create-new-terms-rule, *New terms*>>: Generates an alert for each new term detected in source documents within a specified time range. You can also detect a combination of up to three new terms (for example, a `host.ip` and `host.id` that have never been observed together before).
 
+* <<create-esql-rule, *ES|QL*>>: Uses the Elasticsearch Query Language (ES|QL) to find events in the defined data sources and aggregate search results. 
++
+preview::[]
+
+creates an alert when one or more documents match the rule's ES|QL query.
```
nastasha-solomon, line 49:
```suggestion
```

docs/detections/about-rules.asciidoc
```diff
@@ -42,6 +42,12 @@ TIP: You can also use value lists as the indicator match index. See <<indicator-
 
 * <<create-new-terms-rule, *New terms*>>: Generates an alert for each new term detected in source documents within a specified time range. You can also detect a combination of up to three new terms (for example, a `host.ip` and `host.id` that have never been observed together before).
 
+* <<create-esql-rule, *ES|QL*>>: Uses the Elasticsearch Query Language (ES|QL) to find events in the defined data sources and aggregate search results.
```
nastasha-solomon, line 45:
```suggestion
* <<create-esql-rule, *ES|QL*>>: Uses the Elasticsearch Query Language (ES|QL) to find events and aggregate search results. 
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,24 +263,32 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
-. To create a rule that uses ES|QL, select **ES|QL**, 
-then write a query. There are two types of ES|QL queries:
-
-.. **Aggregating query**: This is a query that uses the `STATS...BY` grouping commands. Query results cannot be matched with a particular document in Elasticsearch. For example:
+. Select **ES|QL**. 
+. Write an aggregating or non-aggregating type query:
++
+TIP: Click the **ES|QL reference** button [book icon] for more information on ES|QL syntax. 
++
+** **Aggregating query**: Uses {ref}/esql-agg-functions.html[STATS...BY] functions to aggregate data. Results from each query row are not saved to an alert document within Elasticsearch.
```
nastasha-solomon, line 274:
```suggestion
** **Aggregating query**: Uses {ref}/esql-functions-operators.html#esql-agg-functions[`STATS...BY`] functions to aggregate data. Results from each query row are not saved to an alert document within Elasticsearch. 
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,33 +263,37 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
-. To create a rule that uses ES|QL, select **ES|QL**, 
-then write a query. There are two types of ES|QL queries:
-
-.. **Aggregating query**: This is a query that uses the `STATS...BY` grouping commands. Query results cannot be matched with a particular document in Elasticsearch. For example:
+. Select **ES|QL**. 
+. Write an aggregating or non-aggregating type query:
 +
-[esql]
------
+TIP: Click the **ES|QL reference** button [book icon] for more information on ES|QL syntax.
```
vitaliidm, line 272: should we specify button located in ES|QL input or add a screenshot with it?

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,33 +263,37 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
-
-.. **Aggregating query**: This is a query that uses the `STATS...BY` grouping commands. Query results cannot be matched with a particular document in Elasticsearch. For example:
+. Select **ES|QL**. 
+. Write an aggregating or non-aggregating type query:
 +
-[esql]
------
+TIP: Click the **ES|QL reference** button [book icon] for more information on ES|QL syntax. 
++
+** **Aggregating query**: Uses {ref}/esql-functions-operators.html#esql-agg-functions[`STATS...BY`] functions to aggregate alert data. An alert is generated for each segment of the query, but is not associated with a specific source event since aggregating queries process data from multiple alert documents. As a result, alerts from aggregating queries do not contain source event data.
```
vitaliidm, line 274: > An alert is generated for each segment of the query,

What is meant by `each segment of the query`?

vitaliidm: I would try to rephrase it as: Generated alerts contain only fields returned by query, without mentioning source event documents. Let's leave it for non-aggregating query only,

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,33 +263,37 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+[source,esql]
+----
 FROM logs*
 | STATS count = COUNT(host.name) BY host.name
 | SORT host.name
------
+----
 
-.. **Non-aggregating query**: This is a query that _does not_ use the `STATS...BY` grouping commands. Each row in the query results can be tracked to a source document in Elasticsearch. 
+** **Non-aggregating query**: Doesn't use `STATS...BY` functions and doesn't aggregate alert data. An alert is generated for each segment of the query, and each alert is associated with a source event. As a result, alerts from non-aggregating queries contain source event data.
```
vitaliidm, line 285: > functions and doesn't aggregate alert data

```suggestion
** **Non-aggregating query**: Doesn't use `STATS...BY` functions and doesn't aggregate source data. An alert is generated for each row in the result, and each alert can be linked to a source event through [metadata] operator. As a result, alerts from non-aggregating queries contain source event data merged with fields returned by query request
```

vitaliidm (commented):
I will continue review tomorrow  and will add some query example to underscore points

docs/detections/rules-ui-create.asciidoc
vitaliidm: row, column, table - something described in our tech docs: https://esql.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/esql.html
So, we can use it our advantage and add link to it in our doc.
And we can put in common area of ES|QL section: each row become an alert. That is cornerstone for how this rule type works

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Yep! Added these deets in [ab28d69](https://github.com/elastic/security-docs/pull/4062/commits/ab28d69cc19c54046fda4c1e54df7dee78bf0ab6).

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: @vitaliidm to clarify, are you suggesting that the second and third sentences be revised like so:

```suggestion
** **Aggregating query**: Uses {ref}/esql-functions-operators.html#esql-agg-functions[`STATS...BY`] functions to aggregate alert data. An alert is generated for each row in the result table, and alert documents only contain fields returned by the query.  
```

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Would you mind elaborating on the last part of your comment? I'm not clear on whether you're suggesting that we explain how the ES|QL rule generates alerts in these docs, or in the [core ES|QL docs](https://www.elastic.co/guide/en/elasticsearch/reference/master/esql.html).

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,33 +263,37 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
-[esql]
+[source,esql]
 -----
 FROM logs* [metadata _id, _index, _version]
-| WHERE event.id == "test"
+| WHERE event.id == "8a4f500d"
 | LIMIT 10
 -----
-+
-Ensure, metadata properties `id`, `_index`, `_version` are carried over through pipe operators.
```
vitaliidm, line 297:
```suggestion
WARNING:  Following examples drop _id field and are not valid for non-aggregating query

In this example command KEEP omits _id property, so it won't be returned in the result table
+
[source,esql]
-----
FROM logs* [metadata _id, _index, _version]
| WHERE event.category == "process"  AND event.id == "8a4f500d"
| KEEP event.*
| LIMIT 10

-----


In the next example command DROP explicitly omits  _id property, so it won't be returned in the result table
+
[source,esql]
-----
FROM logs* [metadata _id, _index, _version]
| WHERE event.category == "process"  AND event.id == "8a4f500d"
| DROP _id
| LIMIT 10

-----
```

vitaliidm: added 2 examples to elaborate what dropping _id means

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,33 +263,37 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
-IMPORTANT: This is a placeholder for future documentation. The following content is incomplete.
+preview::[]
 
 . Go to *Rules* -> *Detection rules (SIEM)* -> *Create new rule*. The *Create new rule* page displays.
-. To create a rule that uses ES|QL, select **ES|QL**, 
-then write a query. There are two types of ES|QL queries:
-
-.. **Aggregating query**: This is a query that uses the `STATS...BY` grouping commands. Query results cannot be matched with a particular document in Elasticsearch. For example:
+. Select **ES|QL**. 
+. Write an aggregating or non-aggregating type query:
```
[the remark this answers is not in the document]
nastasha-solomon: Yeah, I was thinking about making this more general too. Just in case we want to specify more query types in the future. How about something like this:

```suggestion
. Write an ES|QL query. You can create any of the following query types:
```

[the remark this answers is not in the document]
nastasha-solomon: Resolving because I responded [here](https://github.com/elastic/security-docs/pull/4062#discussion_r1371772895).

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,33 +263,37 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
++
+** **Aggregating query**: Uses {ref}/esql-functions-operators.html#esql-agg-functions[`STATS...BY`] functions to aggregate alert data. An alert is generated for each segment of the query, but is not associated with a specific source event since aggregating queries process data from multiple alert documents. As a result, alerts from aggregating queries do not contain source event data.
++
+Example aggregating query: 
++
+[source,esql]
+----
 FROM logs*
 | STATS count = COUNT(host.name) BY host.name
 | SORT host.name
```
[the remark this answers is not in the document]
nastasha-solomon: @vitaliidm I was thinking we could frame these examples in a more descriptive way. Maybe something like:

For example, if you wanted to create an alert for < describe scenario >, you can write a query like this:

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: @vitaliidm in addition to explaining how ES|QL alerts are generated, do you think we need to mention any best practices for writing ES|QL rule queries? For example, if users drop certain fields in the query, will those fields be omitted from the alert document? Also, should they include certain fields to ensure they can use specific functionality in the alert details flyout, like the [Entities](https://www.elastic.co/guide/en/security/master/view-alert-details.html#entities-overview) overview?

docs/detections/rules-ui-create.asciidoc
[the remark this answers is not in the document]
vitaliidm: I don't think we need event this: `You can create any of the following query types:`
Just ` Write an ES|QL query`

docs/detections/rules-ui-create.asciidoc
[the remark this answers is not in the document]
vitaliidm: We can describe this query as: 

> Rule with the following query generates alert for each host.name that encountered more than 20 times during rule interval.

For more production oriented rules, I think we would need to contact TRADE team. But, it can be very complex, bussiness specific rule. I a m not sure we want to add it in our docs, that should have simple enough examples to give basic understanding how rules work

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Oh yeah, I'm not suggesting that we go through all of our features to find fields that users need to keep or return. I'm more so asking if we need to highlight detail because I feel like it's pretty obscured and can affect users's ability to use certain functionality in the flyout or elsewhere.

docs/detections/rules-ui-create.asciidoc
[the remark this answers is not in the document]
nastasha-solomon: Thanks for the description! That was the main item I needed for this part :)

docs/detections/rules-ui-create.asciidoc
vitaliidm: It's difficult to answer, it would depend which flyout(or smth else) functionality we refer. 

As I mentioned  I think it is sufficient enough to point, source document will be merged to alert for non-aggregatin query. Depends on it, user would be able to use other Security Solution features.

I think scope of features too big to mention each single one

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,76 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+===== Non-aggregating query 
+A non-aggregating query doesn't use `STATS...BY` functions and doesn't aggregate source data. Generated alerts contain source event data and fields returned by the ES|QL query. 
+
+If you create a non-aggregating query, you also have to manually turn on alert deduplication and link source events to alerts. To do this, you must add the `[metadata _id, _index, _version]` operator after the `FROM` source command. For example:  
+
+[source,esql]
+-----
+FROM logs* [metadata _id, _index, _version]
+| WHERE event.id == "8a4f500d"
+| LIMIT 10
```
vitaliidm, line 314: Just realised, LIMIT over rule max_signals won't have any effect on number of generated alerts. It will always pick smaller between:
`LIMIT` in ES|QL query and rule [max_signals](https://www.elastic.co/guide/en/security/current/rules-api-create.html#opt-fields-all) in rule's configuration

nastasha-solomon: This is a nice catch, and a good "gotcha" to note in the rule docs. This information applies to ES|QL rule queries in general, so it might make sense to place it under step 2 [here](https://github.com/elastic/security-docs/pull/4062/files#diff-bb6609d5341af025e47b175f8432a1ce9dd90d3c93d656dbc388c599b120924dR273). In the next iteration of these docs, maybe we can move it to a new section/page that provides additional query guidance and examples. 

It might also be a good idea to mention this information about the `LIMIT` and `max_signals` params in the `max_signals` [API docs](https://www.elastic.co/guide/en/security/current/rules-api-create.html#opt-fields-all). Just so there's coverage in both places and users see if whether they create an ES|QL rule from the UI or API. What do you think, @vitaliidm ?

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,76 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+----
 FROM logs*
-| STATS count = COUNT(host.name) BY host.name
-| SORT host.name
+| STATS host_count = COUNT(host.name) BY host.name
+| SORT host_count DESC
+| WHERE host_count > 20
+----
+
+This query generates an alert whenever the `host.name` field is encountered more than 20 times during the specified rule interval.
```
nastasha-solomon, line 301: @vitaliidm it might be helpful to break this example down a bit more so user can see where the aggregation happens. Would you mind verifying that the following information is correct:

```suggestion

- This query starts by querying logs from indices that match the pattern `logs-*`. 
- The query then aggregates the count of events by `host.name`.
- Next, it sorts the result by `host_count` in descending order.
- Then, it filters events where the `host_count` field is encountered more than 20 times during the specified rule interval.
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,76 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+
+[float]
+[[esql-non-agg-query]]
+===== Non-aggregating query 
+A non-aggregating query doesn't use `STATS...BY` functions and doesn't aggregate source data. Generated alerts contain source event data and fields returned by the ES|QL query. 
+
```
nastasha-solomon, line 307: Might be useful to add an example here as well, to keep with the structure of the previous section. Would you mind verifying this as well, @vitaliidm:

```suggestion
Here is an example non-aggregating query:

[source,esql]
-----
FROM logs* [metadata _id, _index, _version]
| WHERE event.category == "process"  AND event.id == "8a4f500d"
| KEEP event.*
| LIMIT 10
-----

- This query starts by querying logs from indices that match the pattern `logs-*`. The [metadata _id, _index, _version] operator is added to turn on alert deduplication.  
- Next, the query filters events where the `event.category` is a process and the `event.id` is `8a4f500d`. 
- It keeps the selected fields `event.*`.
- Then, it limits the output to the top 10 results.
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,76 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+If you create a non-aggregating query, you also have to manually turn on alert deduplication and link source events to alerts. To do this, you must add the `[metadata _id, _index, _version]` operator after the `FROM` source command. For example:
```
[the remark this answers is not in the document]
nastasha-solomon: Will be using this version, which we reviewed in Slack.

```suggestion
To deduplicates alerts, the query needs access to the _id, _index, and _version metadata fields within the queried source event documents.  You can allow this by adding the [metadata _id, _index, _version] operator after the FROM source command, for example:

[source,esql]
-----
FROM logs* [metadata _id, _index, _version]
| WHERE event.category == "process"  AND event.id == "8a4f500d"
| LIMIT 10
Once the metadata fields are provided, a unique alert ID is created for each alert generated by the query.

```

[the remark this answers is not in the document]
nastasha-solomon: Both of these were added in [6ff4a1b](https://github.com/elastic/security-docs/pull/4062/commits/6ff4a1b530582ea52d354361a545455feb0d9388).

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,80 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+This query generates an alert whenever the `host.name` field is encountered more than 20 times during the specified rule interval.
+
+NOTE: Aggregating queries might create duplicate alerts if events that occurred in a rule's additional look-back time are aggregated during the rule's current and previous rule execution cycles.
+
+[float]
+[[esql-non-agg-query]]
+===== Non-aggregating query 
+A non-aggregating query doesn't use `STATS...BY` functions and doesn't aggregate source data. Generated alerts contain source event data and fields returned by the ES|QL query. 
+
+To deduplicates alerts, the query needs access to the `_id`, `_index, and `_version` metadata fields within the queried source event documents.  You can allow this by adding the `[metadata _id, _index, _version]` operator after the FROM source command, for example:
```
nastasha-solomon, line 310:
```suggestion
To deduplicates alerts, the query needs access to the `_id`, `_index`, and `_version` metadata fields within the queried source event documents. You can allow this by adding the `[metadata _id, _index, _version]` operator after the FROM source command, for example:
```

docs/detections/rules-ui-create.asciidoc
vitaliidm: I would phrase it as to allow deduplication, instead of "turn on".

Btw, this query is no valid, as it drops `_id` field. Should not it be a valid one?

nastasha-solomon: Ah yes, good catch. I'll remove that line from the query. How's this:

```suggestion
Here is an example non-aggregating query:
[source,esql]
-----
FROM logs* [metadata _id, _index, _version]
| WHERE event.category == "process"  AND event.id == "8a4f500d"
| LIMIT 10
-----
- This query starts by querying logs from indices that match the pattern `logs-*`. The `[metadata _id, _index, _version]` operator allows alert deduplication.  
- Next, the query filters events where the `event.category` is a process and the `event.id` is `8a4f500d`. 
- Then, it limits the output to the top 10 results.
```

docs/detections/rules-ui-create.asciidoc
vitaliidm: looks good. thansl

docs/detections/rules-ui-create.asciidoc
vitaliidm: Let's move it to step 2.

> It might also be a good idea to mention this information about the LIMIT and max_signals params in the max_signals [API docs](https://www.elastic.co/guide/en/security/current/rules-api-create.html#opt-fields-all). Just so there's coverage in both places and users see if whether they create an ES|QL rule from the UI or API. What do you think, @vitaliidm ?

I don't think we need to add it there. It's only related to ES|QL query.

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,92 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+| SORT host_count DESC
+| WHERE host_count > 20
+----
+
+- This query starts by searching logs from indices that match the pattern `logs-*`. 
+- The query then aggregates the count of events by `host.name`.
+- Next, it sorts the result by `host_count` in descending order.
+- Then, it filters events where the `host_count` field is encountered more than 20 times during the specified rule interval.
+
+NOTE: Aggregating queries might create duplicate alerts if events that occurred in a rule's additional look-back time are aggregated during the rule's current and previous rule execution cycles.
```
vitaliidm, line 306:
```suggestion
NOTE: Agrregating query rules may generate duplicate alerts, when events in additional look-back time getting aggregated twice: in current and previous rule executions
```

queries do not create alerts, alerts getting created during rule execution. ES|QL query just return results, from which alerts are generated

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,92 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+
+[float]
+[[esql-non-agg-query]]
+===== Non-aggregating query 
+A non-aggregating query doesn't use `STATS...BY` functions and doesn't aggregate source data. Generated alerts contain source event data and fields returned by the ES|QL query. 
+Here is an example non-aggregating query:
```
vitaliidm, line 312:
```suggestion
Here is an example of non-aggregating query:
```

nastasha-solomon: We can go with the longer version of this if that's more clear. 

```suggestion
Here is an example of a non-aggregating query:
```

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Oh yeah, I can see how this could be interpreted as ES|QL queries generating alerts. I'm not sure that we want to use the term "agrregating query rules" though. It could be read as a new/different rule type. How about this:

```suggestion
NOTE: Rules that use aggregating queries might create duplicate alerts. This can happen  when events that occur in the additional look-back time are aggregated twice: once in the current rule execution and once in the a previous rule execution.
```

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Quick update: requested language from Vitalii. Am keeping this thread open in the meantime.

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: @vitaliidm I'm merging this for now, but we can revisit it in a couple weeks if needs to adjusted for technical accuracy.

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+- Then, it limits the output to the top 10 results.
 
-.. **Non-aggregating query**: This is a query that _does not_ use the `STATS...BY` grouping commands. Each row in the query results can be tracked to a source document in Elasticsearch. 
-+
-For this type of query, use the operator `[metadata _id, _index, _version]` after defining the index source. This will allow for alerts to be deduplicated and linked to the source documents. For example:
-+
-[esql]
+[float]
+[[esql-non-agg-query-dedupe]]
+===== Deduplicating alerts for rules that use non-aggregating queries
```
nastasha-solomon, line 326:
```suggestion
===== Deduplicating alerts created by rules using non-aggregating queries
```

docs/detections/rules-ui-create.asciidoc
vitaliidm: agree, on aggr rquery rule terminlogy

vitaliidm (approved):
@nastasha-solomon 
Doc page LGTM, thanks a lot for the effort put to make it happen.

The last bit to add, as we discussed on slack, is about limit and max_signals

> When defining the rule’s ES|QL query, there’s no impact to the number of alerts generated if the LIMIT value is higher the rule’s max_signals value.
If the LIMIT value is different from the max_signals value, the rule will choose the lower number and generate alerts based on that.

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Chatted with Vitalii over Slack, and will need to add the following information:
- LIMIT defines how many rows will be returned, and because of this, the max number alerts per rule execution (`max_signals`)cannot be greater than what's defined in LIMIT.
- If LIMIT isn't present, `max_signals` will be taken in account.

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
 
-. Go to *Rules* -> *Detection rules (SIEM)* -> *Create new rule*. The *Create new rule* page displays.
-. To create a rule that uses ES|QL, select **ES|QL**, 
-then write a query. There are two types of ES|QL queries:
+Use {ref}/esql.html[ES|QL] to query your source events and aggregate event data. Query results are returned in a table with rows and columns. Each row becomes an alert.
+
+To create an ES|QL rule:
 
-.. **Aggregating query**: This is a query that uses the `STATS...BY` grouping commands. Query results cannot be matched with a particular document in Elasticsearch. For example:
+. Go to *Rules* -> *Detection rules (SIEM)* -> *Create new rule*. The *Create new rule* page displays.
```
benironside, line 272:
```suggestion
. Go to *Rules* -> *Detection rules (SIEM)* -> *Create new rule*. The *Create new rule* page appears.
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+. Select **ES|QL**, then write a <<esql-rule-query-types,query>>.
```
benironside, line 273:
```suggestion
. Select **ES|QL**, then write a <<esql-rule-query-types,query>>. 
```
Not sure the link is necessary here since it only jumps a few lines. Maybe it's just me but following it was a little disorienting

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+[[esql-rule-query-types]]
+==== ES|QL query types 
+
+ES|QL rule queries are loosely categorized into two types: aggregating and non-aggregating. 
+
+[float]
+[[esql-agg-query]]
+===== Aggregating query 
+
+An aggregating query uses {ref}/esql-functions-operators.html#esql-agg-functions[`STATS...BY`] functions to aggregate source event data. Generated alerts only contain the fields returned by the query.
```
benironside, line 289:
```suggestion
Aggregating queries use {ref}/esql-functions-operators.html#esql-agg-functions[`STATS...BY`] functions to aggregate source event data. The alerts they generate only contain the fields returned by the query.
```
If this is accurate I think it's a bit more clear, but I could be wrong!

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
-| SORT host.name
+| STATS host_count = COUNT(host.name) BY host.name
+| SORT host_count DESC
+| WHERE host_count > 20
+----
+
+- This query starts by searching logs from indices that match the pattern `logs-*`. 
+- The query then aggregates the count of events by `host.name`.
+- Next, it sorts the result by `host_count` in descending order.
+- Then, it filters events where the `host_count` field is encountered more than 20 times during the specified rule interval.
```
benironside, line 304:
```suggestion
- Then, it filters for events where the `host_count` field appears more than 20 times during the specified rule interval.
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+
+NOTE: Rules that use aggregating queries might create duplicate alerts. This can happen  when events that occur in the additional look-back time are aggregated twice: once in the current rule execution and once in the a previous rule execution.
```
benironside, line 306:
```suggestion
NOTE: Rules that use aggregating queries might create duplicate alerts. This can happen  when events that occur in the additional look-back time are aggregated both in the current rule execution and in a previous rule execution.
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+
+[float]
+[[esql-non-agg-query]]
+===== Non-aggregating query 
+A non-aggregating query doesn't use `STATS...BY` functions and doesn't aggregate source event data. Generated alerts contain source event data and fields returned by the ES|QL query.
```
benironside, line 311:
```suggestion
Non-aggregating queries don't use `STATS...BY` functions and don't aggregate source event data. The alerts they generate contain source event data and fields returned by the ES|QL query. 
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+- Then, it filters events where the `host_count` field is encountered more than 20 times during the specified rule interval.
+
+NOTE: Rules that use aggregating queries might create duplicate alerts. This can happen  when events that occur in the additional look-back time are aggregated twice: once in the current rule execution and once in the a previous rule execution.
+
+[float]
+[[esql-non-agg-query]]
+===== Non-aggregating query 
+A non-aggregating query doesn't use `STATS...BY` functions and doesn't aggregate source event data. Generated alerts contain source event data and fields returned by the ES|QL query. 
+
+Here is an example of a non-aggregating query:
```
benironside, line 313:
```suggestion
Here is an example non-aggregating query:
```
Changed to match aggregating query example

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+[source,esql]
+-----
+FROM logs* [metadata _id, _index, _version]
```
benironside, line 316:
```suggestion
FROM logs* [metadata _id, _index, _version]
```
Should this and the previous example say `logs-*` instead of `logs*` ? The description below says `logs-*` and that's what I would expect

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+| WHERE event.category == "process"  AND event.id == "8a4f500d"
+| LIMIT 10
 -----
+- This query starts by querying logs from indices that match the pattern `logs-*`. The `[metadata _id, _index, _version]` operator allows <<esql-non-agg-query-dedupe,alert deduplication>>.  
+- Next, the query filters events where the `event.category` is a process and the `event.id` is `8a4f500d`. 
+- Then, it limits the output to the top 10 results.
```
benironside, line 322: top 10 results organized by what?

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
-.. **Non-aggregating query**: This is a query that _does not_ use the `STATS...BY` grouping commands. Each row in the query results can be tracked to a source document in Elasticsearch. 
-+
-For this type of query, use the operator `[metadata _id, _index, _version]` after defining the index source. This will allow for alerts to be deduplicated and linked to the source documents. For example:
-+
-[esql]
+[float]
+[[esql-non-agg-query-dedupe]]
+===== Turn on alert deduplication for rules using non-aggregating queries
+
+To deduplicate alerts, the query needs access to the `_id`, `_index`, and `_version` metadata fields within the queried source event documents. You can allow this by adding the `[metadata _id, _index, _version]` operator after the `FROM` source command, for example:
```
benironside, line 328:
```suggestion
To deduplicate alerts, a query needs access to the `_id`, `_index`, and `_version` metadata fields of the queried source event documents. You can allow this by adding the `[metadata _id, _index, _version]` operator after the `FROM` source command, for example:
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
 FROM logs* [metadata _id, _index, _version]
-| WHERE event.id == "test"
+| WHERE event.category == "process"  AND event.id == "8a4f500d"
 | LIMIT 10
 -----
-+
-Ensure, metadata properties `id`, `_index`, `_version` are carried over through pipe operators.
 
-. Click *Continue* to <<rule-ui-basic-params, configure basic rule settings>>.
+Once the metadata fields are provided, a unique alert ID is created for each alert generated by the query.
```
benironside, line 337:
```suggestion
When those metadata fields are provided, unique alert IDs are created for each alert generated by the query.
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+
+As you continue developing the query, make sure you don't {ref}/esql-commands.html#esql-drop[`DROP`] or filter out the `_id`, `_index`, or `_version` metadata fields.
```
benironside, line 339:
```suggestion
When developing the query, make sure you don't {ref}/esql-commands.html#esql-drop[`DROP`] or filter out the `_id`, `_index`, or `_version` metadata fields. 
```

docs/detections/rules-ui-create.asciidoc
benironside: This applies to all the examples below as well

docs/detections/rules-ui-create.asciidoc
```diff
@@ -263,35 +263,100 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
 -----
-+
-Ensure, metadata properties `id`, `_index`, `_version` are carried over through pipe operators.
 
-. Click *Continue* to <<rule-ui-basic-params, configure basic rule settings>>.
+Once the metadata fields are provided, a unique alert ID is created for each alert generated by the query.
+
+As you continue developing the query, make sure you don't {ref}/esql-commands.html#esql-drop[`DROP`] or filter out the `_id`, `_index`, or `_version` metadata fields. 
+
+Here is an example of an invalid query that uses the `DROP` command to omit the `_id` property from the results table:
```
benironside, line 341: Are the examples invalid, or would they just fail to deduplicate alerts?

benironside (commented):
Awesome work! Congratulations. Left some suggestions for your consideration, I hope they are helpful!

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Yeah, I've gone back and forth on this too. Other than implying that there's more to learn, it's not adding much value 😅

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: My understanding is that they'd fail to dedupe alerts. Is this version more clear?

```suggestion
Here is an example of a query that fails to deduplicate alerts. It uses the `DROP` command to omit the `_id` property from the results table:
```

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Good catch! Fixed in [cbc6444](https://github.com/elastic/security-docs/pull/4062/commits/cbc64440611a121f93d40cb3ad7d1b4f4c33dc10).

natasha-moore-elastic (commented):
Looks great! 🚀 Just had one question on the ES|QL rule type description.

docs/detections/about-rules.asciidoc
```diff
@@ -42,6 +42,10 @@ TIP: You can also use value lists as the indicator match index. See <<indicator-
 
 * <<create-new-terms-rule, *New terms*>>: Generates an alert for each new term detected in source documents within a specified time range. You can also detect a combination of up to three new terms (for example, a `host.ip` and `host.id` that have never been observed together before).
 
+* <<create-esql-rule, *ES|QL*>>: Uses an {ref}/esql.html[Elasticsearch Query Language (ES|QL)] query to search your events and aggregate source event data.
```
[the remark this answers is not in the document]
nastasha-solomon: Yeah, good point! The current description isn't all that informative if you're wondering how ES|QL rules create alerts. Initially, I had considered writing a general statement like "creates and alert when the query is met", but I decided against it because I didn't think it was descriptive enough. I go into more detail about how alerts are generated in the actual steps though, so writing high-level explanation is probably fine here.

```suggestion
* <<create-esql-rule, *ES|QL*>>: Searches the defined indices and creates an alert when results match an {ref}/esql.html[Elasticsearch Query Language (ES|QL)] query.
```

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Quick update: I'll address this once have more info on [this](https://github.com/elastic/security-docs/pull/4065#discussion_r1377546493).

paulewing (approved):
Looks good to me. Thanks!

docs/detections/about-rules.asciidoc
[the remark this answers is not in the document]
natasha-moore-elastic: Sounds good, thanks for the context!

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Resolving this as discussed in Slack.

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: From a technical perspective, it might be more accurate to say that the rule generates alerts -- at least that's what I'm gathering from @vitaliidm's explanation [here](https://github.com/elastic/security-docs/pull/4062#discussion_r1374795992). 

How does this look, @benironside:
```suggestion
Aggregating queries use {ref}/esql-functions-operators.html#esql-agg-functions[`STATS...BY`] functions to aggregate source event data. Alerts generated by the {esql} rule only contain the fields returned by the query.
```

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Same thing that I mentioned [here](https://github.com/elastic/security-docs/pull/4062#discussion_r1378067372):

Non-aggregating queries don't use `STATS...BY` functions and don't aggregate source event data. Alerts generated by the {esql} rule only contain the fields returned by the query.

docs/detections/rules-ui-create.asciidoc
benironside: Yes 100%. Looks great!

benironside: Yes 100%. Looks great!

docs/detections/rules-ui-create.asciidoc
benironside:
```suggestion
Aggregating queries use {ref}/esql-functions-operators.html#esql-agg-functions[`STATS...BY`] functions to aggregate source event data. Alerts generated by an {esql} rule only contain the fields returned by its query.
```

benironside: What about this?

docs/detections/rules-ui-create.asciidoc
```diff
@@ -261,37 +261,102 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+- The query then aggregates the count of events by `host.name`.
+- Next, it sorts the result by `host_count` in descending order.
+- Then, it filters for events where the `host_count` field appears more than 20 times during the specified rule interval.
+
+NOTE: Rules that use aggregating queries might create duplicate alerts. This can happen  when events that occur in the additional look-back time are aggregated both in the current rule execution and in a previous rule execution.
+
+[float]
+[[esql-non-agg-query]]
+===== Non-aggregating query 
+A non-aggregating query doesn't use `STATS...BY` functions and doesn't aggregate source event data. Generated alerts contain source event data and fields returned by the {esql} query.
```
nastasha-solomon, line 311:
```suggestion
A non-aggregating query doesn't use `STATS...BY` functions and doesn't aggregate source event data. Generated alerts contain source event data and fields returned by the {esql} query. 

Non-aggregating queries doesn't use `STATS...BY` functions and doesn't aggregate source event data. Alerts generated by an {esql} rule with a non-aggregating query only contain the fields returned by the query.
```

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Resolved in [c5e95f5](https://github.com/elastic/security-docs/pull/4062/commits/c5e95f5165a7816ca90f7ac80a3870e14fb89955).

docs/detections/rules-ui-create.asciidoc
```diff
@@ -261,37 +261,102 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+- The query then aggregates the count of events by `host.name`.
+- Next, it sorts the result by `host_count` in descending order.
+- Then, it filters for events where the `host_count` field appears more than 20 times during the specified rule interval.
+
+NOTE: Rules that use aggregating queries might create duplicate alerts. This can happen  when events that occur in the additional look-back time are aggregated both in the current rule execution and in a previous rule execution.
+
+[float]
+[[esql-non-agg-query]]
+===== Non-aggregating query 
+ Non-aggregating queries doesn't use `STATS...BY` functions and doesn't aggregate source event data. Alerts generated by an {esql} rule with a non-aggregating query only contain the fields returned by the query.
```
abdonpijpelink, line 311:
```suggestion
Non-aggregating queries doesn't use `STATS...BY` functions and doesn't aggregate source event data. Alerts generated by an {esql} rule with a non-aggregating query only contain the fields returned by the query.
```
This doesn't render correctly, which I think may be caused by this space.

nastasha-solomon: Yeah, I was just about to fix this copy/paste error. Thank you!

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Added in [2d45097](https://github.com/elastic/security-docs/pull/4062/commits/2d450970274daa8dedf92385092f1d24f825615e).

docs/detections/rules-ui-create.asciidoc
```diff
@@ -261,37 +261,104 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+To create an {esql} rule:
 
-.. **Aggregating query**: This is a query that uses the `STATS...BY` grouping commands. Query results cannot be matched with a particular document in Elasticsearch. For example:
+. Go to *Rules* -> *Detection rules (SIEM)* -> *Create new rule*. The *Create new rule* page appears.
+. Select **{esql}**, then write a <<esql-rule-query-types,query>>. 
++
+TIP: Click the help icon (image:images/esql-ref-button.png[Click the ES|QL help icon,20,20]) to open the in-product reference documentation for all {esql} commands and functions.
 +
-[esql]
+NOTE: To avoid conflicts between the {ref}/esql-commands.html#esql-limit[`LIMIT` value] (which defines how many rows are returned in the results table) and the <<opt-fields-all,`max_signals`>> value (which determines the max number of alerts that can be generated per rule execution), make sure the `max_signals` value does not exceed the `LIMIT` value. If you don't define a `LIMIT` value, the rule automatically uses the `max_signals` value.
```
nastasha-solomon, line 277: @vitaliidm let me know if this note needs any tweaks. Thanks in advance for the feedback!

docs/detections/rules-ui-create.asciidoc
```diff
@@ -261,37 +261,109 @@ For example, if a rule has an interval of 5 minutes, no additional look-back tim
...
+Use {ref}/esql.html[{esql}] to query your source events and aggregate event data. Query results are returned in a table with rows and columns. Each row becomes an alert.
+
+To create an {esql} rule:
+
+. Go to *Rules* -> *Detection rules (SIEM)* -> *Create new rule*. The *Create new rule* page appears.
+. Select **{esql}**, then write a <<esql-rule-query-types,query>>. 
++
+TIP: Click the help icon (image:images/esql-ref-button.png[Click the ES|QL help icon,20,20]) to open the in-product reference documentation for all {esql} commands and functions.
++
+[NOTE]
```
nastasha-solomon, line 277: @vitaliidm let me know if this note needs any tweaks. Thanks in advance for the feedback!

nastasha-solomon: Thanks for the prompt feedback, @vitaliidm!

RE: Your first point about `max_signals` not being configurable through the UI:  

The ability to only configure `max_signals` through the API, and documenting this information there was sorta what I was alluding to [here](https://github.com/elastic/security-docs/pull/4062#discussion_r1374571790). It's also why I linked to the `max_signals` docs in the API docs, but it might be too presumptuous to assume that users would make the connection that the `max_signals` field is only configurable via the Create rule API. 

The clearest and most helpful way to doc this information is to put it where users need it, which is in the API docs. I still think the `max_signals` [docs](https://www.elastic.co/guide/en/security/master/rules-api-create.html#opt-fields-all) is the appropriate place to put this information, even though the description for that field applies generally to all rule types. Adding this information as a note shows that there's a caveat to the functionality that users should be aware of when they're working with a specific rule.

nastasha-solomon: @vitaliidm I'll merge this PR and let's continue this convo over in https://github.com/elastic/security-docs/issues/4195.

PR #4067, opened 2023-10-18
[Detection Engine] update lists API

vitaliidm:
Fixes https://github.com/elastic/security-docs/issues/4069

updates list API description with data stream mentions


docs/detections/api/exceptions/lists-index-api-overview.asciidoc
```diff
@@ -1,17 +1,17 @@
 [[lists-index-api-overview]]
-=== Lists index endpoint
+=== Lists data stream(/index) endpoint
```
vitaliidm, line 2: so, the endpoint is `api/lists/index`, but we create data stream. So, I put `/index` in parentheses here. Let me know, if there any better way to express it

natasha-moore-elastic (commented):
Thanks again for making these changes, @vitaliidm! Looks good overall, just a few questions (GH won't let me comment on the specific lines):

1. In `lists-index-api-overview.asciidoc`, should we update the following ’index’ / ‘indices’ references?
    * line 40: "Get index"
    * line 81: "Example response when the indices do not exist"

2. In `detections-req.asciidoc`, should we add “and data streams” on the following lines?

    * line 74:  "The `manage`, `write`,`read`, and `view_index_metadata` index privileges for the following system indices"
    * line 99: "The `manage`, `write`,`read`, and `view_index_metadata` index privileges for the following system indices, where `<space-id>` is the {kib} space name"
    * line 121: "The `maintenance`, `write`,`read`, and `view_index_metadata` index privileges for the following system indices, where `<space-id>` is the {kib} space name"

3. If the endpoint itself is still `api/lists/index`, I’m wondering if we should keep the `lists-index-api-overview.html` page title as **Lists index endpoint**, to avoid confusion?

vitaliidm:
@natasha-moore-elastic 
Thanks for catching missing items.

I have updated first 2 points.
But don't have strong opinion on the third one - once we mentioned everywhere that .lists and .items are data streams, we probably can change that title.

nastasha-solomon:
> If the endpoint itself is still api/lists/index, I’m wondering if we should keep the lists-index-api-overview.html page title as Lists index endpoint, to avoid confusion?

I agree, @natasha-moore-elastic. Another reason to keep the title as **Lists index endpoint**: My _very_ rudimentary understanding of data streams, is that they're made up of indices. If this is the case, the current title (**Lists index endpoint**) is technically correct. 

Also, I'm not sure where this UI text is stored, but I think we also need to update the banner that appears on the Alerts page if users don't have certain privs. This is what I'm talking about, @vitaliidm:


Screenshot 2023-10-25 at 3 31 43 PM

vitaliidm:
@nastasha-solomon , changed back list index endpoint

nastasha-solomon:
@vitaliidm can you elaborate on this:

> It worth to note, it is still missing index priveleges, but for data streams. 

I'm mainly suggesting that we update the **Insufficient banner** text so that it aligns with the changes you made [here](https://github.com/elastic/security-docs/pull/4067/files#diff-7071d3f8f16c85ba42ead73e4dfde38998099dc9cba07ef323600d4da086ff20).

nastasha-solomon (approved):
Looks good to go!

PR #4070, opened 2023-10-18
[DOCS] Add custom fields to cases

lcawl:
Relates to https://github.com/elastic/kibana/issues/160236

This PR updates the [Security Cases documentation](https://www.elastic.co/guide/en/security/master/cases-open-manage.html) with information about how to use custom fields.

It also fixes a link in the email notification steps (per https://github.com/elastic/kibana/pull/168656).

NOTE: This PR includes two less screenshots than I added in https://github.com/elastic/kibana/pull/169057. Those screenshots show where the custom fields appear when you create a new case and edit an existing case. If you want those screenshots here too, just lmk and I can automate and add them as well.


docs/cases/cases-manage.asciidoc
```diff
@@ -37,6 +37,46 @@ NOTE: If you've selected a connector for the case, the case is automatically pus
...
+[role="screenshot"]
+image::images/cases-add-custom-field.png[Add a custom field in case settings]
+// NOTE: This is an autogenerated screenshot. Do not edit it directly.
+--
+
+. Enter a field label.
+
+. Choose a field type: text or toggle.
+
+. If you want the text field to be mandatory in all cases, select *Make this field required*.
```
nastasha-solomon, line 70:
```suggestion
. (Optional) To make the text field mandatory, select *Make this field required*.
```

docs/cases/cases-manage.asciidoc
```diff
@@ -37,6 +37,46 @@ NOTE: If you've selected a connector for the case, the case is automatically pus
...
+
+. Click *Save field*.
+
+You can subsequently remove or edit custom fields on the *Settings* page.
```
nastasha-solomon, line 74: I kinda want to bring the next three lines together so they look less disjointed. Here's an option:

```suggestion
After you create custom fields, they're added to all new and existing cases. Existing cases have null values for new text fields until you set them in. To manage custom fields, go the *Settings* page and remove or edit them.
```

nastasha-solomon (approved):
Looks great! Just left two suggestions for your consideration. Thank you for taking care of this!

docs/cases/cases-manage.asciidoc
lcawl: Fixed in f0bb1b1e12afa0e2583223dce275bf58fce07103, thanks!

PR #4142, opened 2023-10-31
Add .vscode to .gitignore file

joepeeples:
Add `.vscode/` to the .gitignore file for the `security-docs` repo. This allows contributors to customize their local VS Code app's workspaces without Git trying to track those customizations (which are stored within each repo's .vscode subdirectory). 

This is consistent with how many other repos manage their .gitignore files, including Security's own [Serverless docs repo](https://github.com/elastic/staging-serverless-security-docs/blob/main/.gitignore#L2).

We're backporting to current live branches (main, 8.11, and 7.17). Those are the branches we're most likely to be updating in the future. If we need to update other branches, we can handle case-by-case.


jmikell821 (approved):
Not sure if this should be out of `draft` stage yet -- appears so? But LGTM!

joepeeples:
@jmikell821 It's still in draft — I need to check with some of the other teams to see if there's anything else we need to do for this, and how to handle backports. I was planning to get back to this after the 8.11 release crunch.

joepeeples:
Updating backport labels to focus on current live branches (`main`, `8.11`, and `7.17`). Those are the branches we're most likely to be updating. If we need to update other branches, we can handle case-by-case.

PR #4228, opened 2023-11-08
[Security Solution] Add known issue for date selection in Protection Updates

kevinlog:
We discovered an issue in the Protection Updates feature released in `8.11.0` that we will fix for `8.11.1`.  We'd like to add another entry to known issues for `8.11.0` before the fix goes out in the next patch.

The issue is that if the user picks the current date, it's possible that we haven't released a protection update yet for that date, so the Agent will go to an Unhealthy state.  The workaround is to pick a previous date like the day before.  The upcoming fix will simply disallow users from selecting the current date from the UI to reduce the user confusion

Related bug ticket: https://github.com/elastic/kibana/issues/170847


joepeeples (commented):
A few wording & syntax suggestions — thanks @kevinlog for creating the PR!

kevinlog:
Thanks @joepeeples wording sounds good to me - I accepted and committed the change

nastasha-solomon (approved):

Issue #4195, opened 2023-11-06
Expand on ES|QL rule docs


nastasha-solomon:
## Description 

It'd be useful to include the following best practices, limitations, and gotchas in the ES|QL rule docs:

From @vitaliidm:
- If the `LIMIT` value is greater than `max_signals`, the number of generated alerts will be capped by `max_signals`. For example:
    - When `LIMIT` is lesser than `max signals`, number of alerts will be equal to `LIMIT`
    - When `LIMIT` is greater than `max signals`, number of alerts will be equal to `max_signals`

From @paulewing:
- If you create a new field via ES|QL, you will not be able to filter/search for that field on the alerts page (or any security page) unless you create a runtime field for the new field name. This is possible via our field browser, but could be a nice note to the user.
- Similar to above, currently its not possible to add custom highlighted fields for a new field created by ES|QL. A runtime field does NOT solve this, its just a limitation for now.
- For aggregation ES|QL queries (e.g. STATS), it is beneficial for the user to keep fields that they want to search/filter on in the alerts page (e.g. host.name, user.name, process.name, etc.)

## Notes

- [Thread](https://github.com/elastic/security-docs/pull/4062#discussion_r1380804828) with more info on the `LIMIT` vs `max_signals` case. 
- Might make sense to add a new section, or two, for this information (e.g., "Limitations" and "Best practices").

PR #4235, opened 2023-11-10
Expand on ES|QL rule docs

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/4195.

Previews:
- [Query design considerations](https://security-docs_4235.docs-preview.app.elstc.co/guide/en/security/master/rules-ui-create.html#esql-query-design)
- [ES|QL rule limitations](https://security-docs_4235.docs-preview.app.elstc.co/guide/en/security/master/rules-ui-create.html#esql-rule-limitations)


docs/detections/rules-ui-create.asciidoc
```diff
@@ -358,6 +360,30 @@ FROM logs-* [metadata _id, _index, _version]
 | LIMIT 10
 -----
 
+[float]
+[[esql-query-design]]
+===== Query design considerations 
+
+When writing your query, consider the following:
+
+- The {ref}/esql-commands.html#esql-limit[`LIMIT`] command specifies the number of rows an {esql} query returns and the number of alerts created per rule execution. Similarly, the <<opt-fields-all,`max_signals`>> setting specifies how many alerts a rule can create every time it runs.
```
vitaliidm, line 369:
```suggestion
- The {ref}/esql-commands.html#esql-limit[`LIMIT`] command specifies the number of rows an {esql} query returns and the number of alerts created per rule execution. Similarly, the <<opt-fields-all,`max_signals`>> setting specifies the max number of alerts a rule can create every time it runs.
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -358,6 +360,30 @@ FROM logs-* [metadata _id, _index, _version]
...
++ 
+If the `LIMIT` value is lower than the `max_signals` value, the rule uses the `LIMIT` value. If the `LIMIT` is higher than `max_signals`, the rule uses `max_signals`.
```
vitaliidm, line 371:
```suggestion
If the `LIMIT` value is lower than the `max_signals` value, the rule uses the `LIMIT` value. If the `LIMIT` is higher than `max_signals`, the rule uses `max_signals` to determine the max number of generated alerts. 
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -358,6 +360,30 @@ FROM logs-* [metadata _id, _index, _version]
...
+When writing your query, consider the following:
+
+- The {ref}/esql-commands.html#esql-limit[`LIMIT`] command specifies the number of rows an {esql} query returns and the number of alerts created per rule execution. Similarly, the <<opt-fields-all,`max_signals`>> setting specifies how many alerts a rule can create every time it runs.
++ 
+If the `LIMIT` value is lower than the `max_signals` value, the rule uses the `LIMIT` value. If the `LIMIT` is higher than `max_signals`, the rule uses `max_signals`. 
++
+NOTE: The `max_signals` default value is 100. You can modify it using the <<rules-api-create,Create rule API>>. 
++
+
+- When writing an aggregating query, use the {ref}/esql-commands.html#esql-keep[`KEEP`] command to keep fields that you want to examine. For example, if you want to search and filter for the `host.name`, `user.name`, and `process.name` fields, keep them in your query.
```
vitaliidm, line 376: This won't work. Aggregating query returns only fields defined after ~BY~ keyword, apart from field that has value of aggregation

https://www.elastic.co/guide/en/elasticsearch/reference/master/esql-commands.html#esql-stats-by

> The STATS ... BY processing command groups rows according to a common value and calculate one or more aggregated values over the grouped rows. If BY is omitted, the output table contains exactly one row with the aggregations applied over the entire dataset.

So users will need to use these fields in `BY` clause of the command

docs/detections/rules-ui-create.asciidoc
```diff
@@ -358,6 +360,30 @@ FROM logs-* [metadata _id, _index, _version]
...
+
+- When writing an aggregating query, use the {ref}/esql-commands.html#esql-keep[`KEEP`] command to keep fields that you want to examine. For example, if you want to search and filter for the `host.name`, `user.name`, and `process.name` fields, keep them in your query. 
+
+[float]
+[[esql-rule-limitations]]
+===== {esql} rule limitations  
+
+The {esql} rule has the following limitations:
+
+- If your query contains new fields that aren't {ecs-ref}[ECS-compliant], the detection engine won't write them to alert documents, and you can't search or filter for them from the Alerts table. As a workaround, create <<runtime-fields,runtime fields>>.
```
vitaliidm, line 384: Detection engine will write the to alerts docs, if these fields do not conflict with ECS mapping. (yes they can conflict and will be omitted. Please refer to this slack thread https://elastic.slack.com/archives/C6E3MTCD7/p1699635283801569)

While they written, they still can't be searched or filtered by

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Small edit:

```suggestion
- The {ref}/esql-commands.html#esql-limit[`LIMIT`] command specifies the number of rows an {esql} query returns and the number of alerts created per rule execution. Similarly, the <<opt-fields-all,`max_signals`>> setting specifies the maximum number of alerts a rule can create every time it runs.
```

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: I'm not sure we need to be this explicit. These sentences are directly after a brief explanation of the `LIMIT` command and `max_signals` setting, so they're presented within that context. If you still think the text you added needs to be included for clarity, what do you think about adding it to the first sentence and adding "value" to the second sentence? For example:

```suggestion
If the `LIMIT` value is lower than the `max_signals` value, the rule uses the `LIMIT` value to determine the maximum number of alerts the rule generates. If the `LIMIT` value is higher than `max_signals` value, the rule uses `max_signals` value. 
```

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: @vitaliidm thanks for the check here - how should this be re-written?

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Thanks for the additional resources, @vitaliidm. If you can point me towards a specific message or summarize the main points that you think are important, that'd help me understand the information faster. 

In terms of updating this bullet, am I understanding correctly that newly-created fields _are_ written to to alert documents as long as they don't conflict with the ECS mappings? If that's the case, what prevents them from being searched or filtered for?

docs/detections/rules-ui-create.asciidoc
vitaliidm: I like the last suggestion, let's go with this one

docs/detections/rules-ui-create.asciidoc
vitaliidm: I would say 
```suggestion
- When writing an aggregating query, if you want to search and filter for the `host.name`, `user.name`, `process.name` fields or any other field, place them after {ref}/esql-commands.html#esql-stats-by[`BY operator`]  command.
```

docs/detections/rules-ui-create.asciidoc
vitaliidm: https://elastic.slack.com/archives/C6E3MTCD7/p1699636456081289?thread_ts=1699635283.801569&cid=C6E3MTCD7

Here is the ticket that describes issue.

> In terms of updating this bullet, am I understanding correctly that newly-created fields are written to to alert documents as long as they don't conflict with the ECS mappings? 

It's correct, For example, you can't write `host: test-1` to alert, because `host` is ECS field and is object, not string.

> If that's the case, what prevents them from being searched or filtered for?

Alerts index configured in a way it does not index unmapped fields, only ECS. But fields are stored within document and returned in document itself. Refer to value `false` in this section: https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html#dynamic-parameters
Also, something discussed on the last adv corr WG: https://elastic.slack.com/archives/C03E8TR26HE/p1699652543710329

docs/detections/rules-ui-create.asciidoc
vitaliidm: >  For example, placing the `host.name`, `user.name`, `process.name` fields after the BY operator of `STATS...BY` command returns these fields in alert documents, and allows you to search and filter for them from the Alerts table.

nastasha-solomon:
```suggestion
- When writing an aggregating query, follow the {ref}/esql-commands.html#esql-stats-by[`STATS...BY`] command with fields that you want to search and filter for when alerts are created. For example, placing the `host.name`, `user.name`, `process.name` fields after the `BY` operator of the `STATS...BY` command returns these fields in alert documents, and allows you to search and filter for them from the Alerts table. 
```

docs/detections/rules-ui-create.asciidoc
nastasha-solomon: Hey, @vitaliidm - am I understanding correctly that the default alert index doesn't support dynamic mapping, so newly-created fields aren't mapped to alert documents or indexed within the default alert index?

Also, when you say "fields are stored within documents and returned in the document itself", are you saying that newly-created fields are stored in _source_ documents and then returned in the _alert_ document when users search for fields? It'd be helpful to see an example of this if possible. I'm having a hard time visualizing an alert doc that was created from a rule using an aggregating query 😅

nastasha-solomon: Ok, so to summarize, new fields in alert documents are written to to alert documents (as long as they don't conflict with the ECS mappings) but are not dynamically mapped in the alert index. To create that mapping, users must create runtime fields for the new fields. Only then, will they be able to search or filter for the fields in the Alerts table. If this is correct, how does the following revised text look:

```suggestion
- If your {esql} query creates new fields, they won't be mapped to the alerts index (`.alerts-security.alerts-<space-id>`) and you can't search or filter for them from the Alerts table. As a workaround, create <<runtime-fields,runtime fields>> for the new fields.
```

vitaliidm: Maybe we still worth mention about ECS?

> - If your {esql} query creates new non ECS fields, they won't be mapped to the alerts index (`.alerts-security.alerts-<space-id>`) and you can't search or filter for them from the Alerts table. As a workaround, create <<runtime-fields,runtime fields>> for the new fields.

nastasha-solomon: How's this:

```suggestion
- If your {esql} query creates new fields that aren't ECS-compliant, they won't be mapped to the alerts index (`.alerts-security.alerts-<space-id>`) and you can't search or filter for them from the Alerts table. As a workaround, create <<runtime-fields,runtime fields>> for the new fields.
```

nastasha-solomon: Using this version instead, as discussed with @vitaliidm over slack:

```suggestion
- If your {esql} query creates new fields that aren’t part of the ECS schema (non-ECS fields), they won’t be mapped to the alerts index and you can't search or filter for them from the Alerts table. As a workaround, create <<runtime-fields,runtime fields>>. 
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -358,6 +360,30 @@ FROM logs-* [metadata _id, _index, _version]
...
+- When writing an aggregating query, follow the {ref}/esql-commands.html#esql-stats-by[`STATS...BY`] command with fields that you want to search and filter for when alerts are created. For example, placing the `host.name`, `user.name`, `process.name` fields after the `BY` operator of the `STATS...BY` command returns these fields in alert documents, and allows you to search and filter for them from the Alerts table. 
+
+[float]
+[[esql-rule-limitations]]
+===== {esql} rule limitations  
+
+The {esql} rule has the following limitations:
+
+- If your {esql} query creates new fields that aren’t part of the ECS schema (non-ECS fields), they won’t be mapped to the alerts index and you can't search or filter for them from the Alerts table. As a workaround, create <<runtime-fields,runtime fields>>. 
+- Newly-created, non-ECS-compliant fields can't be specified as <<rule-ui-advanced-params,custom highlighted fields>>.
```
nastasha-solomon, line 385: Discussed this with @vitaliidm in Slack, and will need to update it as follows:
```suggestion
- If your {esql} query creates new fields that aren’t in the query’s source index, they can’t be added to the rule’s <<rule-ui-advanced-params,custom highlighted fields>>. There’s no workaround for this.
```

joepeeples (commented):
Added a few small suggested edits, thanks!

Also, sometime soon we might want to think again about splitting up this page into rule types, or something. The ES|QL section alone is getting long, even though this is all necessary info to include. Nothing we need to do now in this PR, of course; looking at the preview just made me realize (again) how long the page is.

docs/detections/rules-ui-create.asciidoc
```diff
@@ -358,6 +360,30 @@ FROM logs-* [metadata _id, _index, _version]
...
+- When writing an aggregating query, follow the {ref}/esql-commands.html#esql-stats-by[`STATS...BY`] command with fields that you want to search and filter for after alerts are created. For example, placing the `host.name`, `user.name`, `process.name` fields after the `BY` operator of the `STATS...BY` command returns these fields in alert documents, and allows you to search and filter for them from the Alerts table. 
+
+[float]
+[[esql-rule-limitations]]
+===== {esql} rule limitations  
+
+The {esql} rule has the following limitations:
+
+- If your {esql} query creates new fields that aren’t part of the ECS schema (non-ECS fields), they won’t be mapped to the alerts index and you can't search or filter for them from the Alerts table. As a workaround, create <<runtime-fields,runtime fields>>. 
+- If your {esql} query creates new fields that aren’t in the query’s source index, they can’t be added to the rule’s <<rule-ui-advanced-params,custom highlighted fields>>. There’s no workaround for this.
```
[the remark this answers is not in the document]
nastasha-solomon: Yeah, I agree that the tone is pretty blunt. I considered removing the sentence, but wasn't sure if omitting it would leave room for speculation. I can't think of an alternative at the moment though, so I'm fine taking it out. 

```suggestion
- If your {esql} query creates new fields that aren’t in the query’s source index, they can’t be added to the rule’s <<rule-ui-advanced-params,custom highlighted fields>>.
```

docs/detections/rules-ui-create.asciidoc
```diff
@@ -358,6 +360,30 @@ FROM logs-* [metadata _id, _index, _version]
...
+When writing your query, consider the following:
+
+- The {ref}/esql-commands.html#esql-limit[`LIMIT`] command specifies the number of rows an {esql} query returns and the number of alerts created per rule execution. Similarly, the <<opt-fields-all,`max_signals`>> setting specifies the maximum number of alerts a rule can create every time it runs.
++ 
+If the `LIMIT` value is lower than the `max_signals` value, the rule uses the `LIMIT` value to determine the maximum number of alerts the rule generates. If the `LIMIT` value is higher than `max_signals` value, the rule uses `max_signals` value. 
++
+NOTE: The `max_signals` default value is 100. You can modify it using the <<rules-api-create,Create rule API>>. 
++
+
+- When writing an aggregating query, follow the {ref}/esql-commands.html#esql-stats-by[`STATS...BY`] command with fields that you want to search and filter for after alerts are created. For example, placing the `host.name`, `user.name`, `process.name` fields after the `BY` operator of the `STATS...BY` command returns these fields in alert documents, and allows you to search and filter for them from the Alerts table.
```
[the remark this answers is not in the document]
nastasha-solomon: All good points - thanks for the tweaks and explanation here!

joepeeples (approved):

PR #4360, opened 2023-12-01
Adds `include_expired_exceptions` to Export exception list API

### LINKED ISSUE #3061: Adding exceptions list export route documentation
## Description 

The documentation seems to be missing for our exception list export which was added (I believe) in [this PR](https://github.com/elastic/kibana/pull/86135) dating back all the way to some `7.x` builds. We should add this in and backport it to the relevant releases. From what I can tell, the route and its query parameters have remained the same since it was written, but we added another query parameter in the [exceptions TTL PR](https://github.com/elastic/kibana/pull/145180) which should be reflected only in the `8.7` docs version

This route itself exports an exception list by providing an exception list ID and the list is exported into an `.ndjson` format. Exception lists consist of two elements - the list itself, and its items. The export file should now contain both these elements, the list followed by its items. The first PR link has examples of exported lists.

The route, following the current docs format, should be `POST <kibana host>:<port>/api/exception_lists/_export`

The query parameters are 
```
{
  id: string // The exception list id
  list_id: string // The exception list list_id
  namespace_type: "agnostic" | "single" // The exception list namespace_type
  
  (for 8.7+ only)
  include_expired_exceptions: (optional) boolean // whether or not to include expired exceptions in the exported list (defaults to true)
}
```

So an example request would be 
```
POST api/exception_lists/_export?id=test_id&list_id=test_list_id&namespace_type=agnostic
```
with the response being an aforementioned `.ndjson` file

natasha-moore-elastic:
Resolves #3061 by adding the `include_expired_exceptions` to versions `8.7` and newer.

Preview: [Export exception list](https://security-docs_4360.docs-preview.app.elstc.co/guide/en/security/master/exceptions-api-export-exception-list.html)

### DIFF
--- a/docs/detections/api/exceptions/api-export-exception-list.asciidoc
+++ b/docs/detections/api/exceptions/api-export-exception-list.asciidoc
@@ -21,6 +21,7 @@ Exports an exception list and its associated items to an `.ndjson` file.
 * `agnostic`: Available in all {kib} spaces.
 
 |No, defaults to `single`.
+|`include_expired_exceptions` |Boolean |Determines whether to include expired exceptions in the exported list. |No, defaults to `true`.
 |==============================================
 
 ===== Example request


benironside (approved):
LGTM!

PR #4462, opened 2023-12-12
Renames internal doc request template to change ToC order

benironside:
Minor change to update the order of the issue templates

[Preview](https://github.com/elastic/security-docs/blob/enhancement-req-yaml-new-title/.github/ISSUE_TEMPLATE/docs-request-internal.yaml) (Unchanged)


joepeeples:
I'm not noticing a difference in the diff, other than adding a space at the end. Will that affect the order?
image

nastasha-solomon (approved):

joepeeples:
> I'm not noticing a difference in the diff, other than adding a space at the end. Will that affect the order?

Nevermind, I just noticed the file name was changed, and I assume that's what determines the order, not the `name` value. FYI, we might also be able to control order and other things by [creating a config.yml file](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser), but it's not necessary here since the file rename does what we need. LGTM!

joepeeples (approved):

Issue #4312, opened 2023-11-27
Bidirectional integration response actions (SentinelOne)


joepeeples:
# Description 

Create docs and docs links for bi-directional integration response actions, per https://github.com/elastic/security-team/issues/7780#issuecomment-1828209705.

# Response action configuration
_From https://github.com/elastic/security-docs/issues/4534, which we closed in favor of this meta-ish issue._

First release of our bidirectional integration work - https://github.com/elastic/security-team/issues/6200

There are a handful of steps users must take to ensure they are able to successfully take action on Sentinel One hosts (for the initial tech preview release in 8.12, the only action is to isolate / release a host). We want to make sure we have a page that walks through the full configuration requirements.

Prerequesites:
* Should have active running S1 agents

High level steps:
1.  Ensure there is an agent policy created (Fleet) that will be responsible for pulling S1 data)
2. Add and configure the Sentinel One integration to policy
3. Create a security detection rule to get elastic alerts for Sentinel One alerts
     * logs-sentinel_one.alert*
5. Create a connector for Sentinel One
   * Needed for manual response actions; automated actions not available yet
   * Create ONLY one connector
   * API token needs to have privileges to perform actions (isolate/release host) 

# Information needed

* Which documentation set does this change impact? (ESS, serverless, or both) — Both
* Feature differences (ESS vs. serverless) — n/a
* ESS release — 8.12.0
* Serverless release — January 16, 2024 (coordinated with ESS release)
* API docs impact — ???
* License/subscription/tier required:
   * ESS — `Enterprise`
   * Serverless — `Endpoint Protection Complete`
* RBAC required:
   * ESS permissions/roles — `all` permissions for connectors & actions (may change in future releases)
   * Serverless preset roles — `SOC manager`, `Endpoint operations analyst` (per [security-team#7911](https://github.com/elastic/security-team/issues/7911))
* Feature flags needed to enable feature — ESS available as of 8.12 BC5. Serverless flags enabled to hide in serverless until ESS release.

---
# Pull requests, related tasks
```[tasklist]
- [x] **Create stub page for docs link: configure response actions (Classic)** — https://github.com/elastic/security-docs/pull/4317
- [x] **Create main docs (Classic)** — https://github.com/elastic/security-docs/pull/4593
- [x] **Update S1 integration docs** — https://github.com/elastic/integrations/pull/8611. ~~🚫 Will also 404 until 8.12 is published, due to cross-docs link.~~
[...]

PR #4593, opened 2024-01-10
Bidirectional integration response actions (SentinelOne) — Classic docs

joepeeples:
Contributes to https://github.com/elastic/security-docs/issues/4312 by adding high-level instructions on configuring SentinelOne's bidirectional response actions to the classic docs.

Preview: [Response actions configuration](https://security-docs_bk_4593.docs-preview.app.elstc.co/guide/en/security/master/response-actions-config.html)

### 🚫 Dependency/Blocker

This PR includes a link (in Step 3) to the [Connectors landing page](https://www.elastic.co/guide/en/kibana/current/action-types.html) in Kibana docs. This needs to be revised to a specific page for the SentinelOne connector docs, which won't exist until https://github.com/elastic/kibana/pull/174696 is merged.

Update: We'll actually go ahead and merge this without revising the link to Kibana's connector docs. We can update it later once https://github.com/elastic/kibana/pull/174696 is merged.


joepeeples:
run docs-build

natasha-moore-elastic (commented):
Looks great! Just a couple of minor suggestions.

docs/management/admin/response-actions-config.asciidoc
```diff
@@ -6,8 +6,68 @@
...
+.. In {kib}, go to **Stack Management** → **Connectors**, then select **Create connector**.
+.. Select the **SentinelOne** connector.
+.. Enter the configuration information:
+   - **Connector name**: A name to identify the connector.
+   - **SentinelOne tenant URL**: The SentinelOne console URL.
+   - **API token**: The SentinelOne API access token you generated previously.
+.. Click **Save**.
+====
+
+. **Create and enable a rule to generate {elastic-sec} alerts.** Create a <<rules-ui-create,detection rule>> to generate {elastic-sec} alerts whenever SentinelOne generates alerts. Use the index pattern `logs-sentinel_one.alert*` when configuring the rule to target the data collected from SentinelOne.
```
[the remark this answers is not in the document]
joepeeples: @natasha-moore-elastic I assume it can be any rule type as long as it searches that index pattern, but maybe we should give more specific guidance here on creating the rule. @paul-tavares @patrykkopycinski WDYT?

docs/management/admin/response-actions-config.asciidoc
```diff
@@ -6,8 +6,68 @@
...
+* {kibana-ref}/kibana-role-management.html#adding_kibana_privileges[{kib} feature privilege]: **Actions and Connectors : All**.
+
+* <<endpoint-management-req,{elastic-sec} feature privileges>>: **All** for any response action features, such as **Host Isolation**, that you want to perform.
+
+* Endpoints must have actively running SentinelOne agents installed.
 --
+
+Configuration requires the following general steps. Expand the steps and follow the links for detailed instructions:
+
+. **Generate an API access token in SentinelOne.** You'll need this token in later steps, and it allows {elastic-sec} to collect data and perform actions in SentinelOne. You must create the token with the SentinelOne permissions required for the actions you want to perform (for example, isolating and releasing an endpoint).
```
paul-tavares, line 33: A few thoughts on here:

I doubt that a customer would use the same API key for both the Integration setup (which enables data to be collected from SentinelOne to Elasticsearch) **and** for the connectors setup (which enables the ability for a user to send response actions to the sentinelOne agents).  In my opinion, I would not advise a customer to enter an API key that can do response actions when setting up the integration in fleet. 

You might want to expand on this a little more to indicate that. So basially:

- for the S1 integration: API key needs to at least be able to read data out of sentinelone's system
- for the connector: the API key need to be able to execute response actions (in this release, `isoalte` and `release` type of response actions)
-

docs/management/admin/response-actions-config.asciidoc
```diff
@@ -6,8 +6,68 @@
...
+[%collapsible]
+====
+.. In {kib}, go to **Integrations**, search for and select **SentinelOne**, then select **Add SentinelOne**.
+.. Configure the integration with an **Integration name** and optional **Description**.
+.. Ensure that **Collect SentinelOne logs via API** is selected, and enter the required **Settings**:
+   - **URL**: The SentinelOne console URL.
+   - **API Token**: The SentinelOne API access token you generated previously.
+.. Scroll down and enter a name for the agent policy in **New agent policy name**. If other agent policies already exist, you can click the **Existing hosts** tab and select an existing policy instead. For more details on {agent} configuration settings, refer to {fleet-guide}/agent-policy.html[{agent} policies].
+.. Click **Save and continue**.
+.. If you need to install {agent} on any SentinelOne-protected hosts, select *Add {agent} to your hosts* and continue with the <<enroll-agent,{agent} installation steps>>.
```
paul-tavares, line 49: Do we already document "agent-less" fleet policies somewhere? 

if not, then I would suggest clarifying here that this integration/fleet agent should be deployed to 1 agent only. That agent will act as the "middle" process in pulling the data out of the SentinelOne management platform and then sending that to elasticsearch, thus it does not need to be deployed to more than 1 host (and likely would never be deployed to a host running the SentinelOne agent)

docs/management/admin/response-actions-config.asciidoc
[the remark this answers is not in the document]
paul-tavares: Yes, it needs to be a `Detection rule (SIEM)` using a `Custom query`.
Also, in addition to to setting the source to (ONLY) `logs-sentinel_one_alert*`, they also need to enter a `custom query` value of `observer.serial_number:*`

See screen capture below

_______

image

paul-tavares (commented):
Hey @joepeeples ,
I did a review in @patrykkopycinski 's absence and left some feedback.

docs/management/admin/response-actions-config.asciidoc
joepeeples: Gotcha, so they'll actually need two separate API keys for SentinelOne, each scoped for least privilege needed for the integration and the connector. I'll push some edits that address this and the other feedback comments; that might cause GitHub to close or hide some of the conversations, we'll see.

docs/management/admin/response-actions-config.asciidoc
paul-tavares: Agent-less policies is not something (I think) that exists yet, but something that has been talked about. The current work-around is to instruct the customer to deploy this fleet policy to a single host using Elastic Defenc (for the reason I highlighted above).

I guess the edit would need to happen on the line with: `...on any SentinelOne-protected hosts...` ⬅️ since this is not really true.

docs/management/admin/response-actions-config.asciidoc
```diff
@@ -6,8 +6,88 @@
...
+. **Generate API access tokens in SentinelOne.** You'll need these tokens in later steps, and they allow {elastic-sec} to collect data and perform actions in SentinelOne. 
++
+.Expand for details
+[%collapsible]
+====
+Create two API tokens in SentinelOne, and give them the least privilege required by the Elastic components that will use them:
+
+- SentinelOne integration: Permission to read SentinelOne data.
+- SentinelOne connector: Permission to perform actions on SentinelOne-protected hosts (for example, isolating and releasing an endpoint).
+
```
paul-tavares, line 42: This is where @patrykkopycinski 's knowledge would be good.

I would say: this Connector API key should have ability to read data (ex. we query for agent details) as well as Isolate/Release), so you may want to expand it a bit more to include something more about being able to also read data

docs/management/admin/response-actions-config.asciidoc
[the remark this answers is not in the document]
joepeeples: Revised in https://github.com/elastic/security-docs/pull/4593/commits/08062e96136b3d4d8b7958372eaec8bad5a541e1

docs/management/admin/response-actions-config.asciidoc
joepeeples: Revised the Agent installation step in https://github.com/elastic/security-docs/pull/4593/commits/ac6b1443f12e18e446d37799406f76d59d8fa99a

paul-tavares (approved):
Thank you for all the changes. 👍

nastasha-solomon (approved):
lgtm!

Issue #4626, opened 2024-01-18
[BUG] Outdated information mentioned in the step 5 for Osquery.


sukhwindersingh-qasource:
### Documentation links

https://www.elastic.co/guide/en/security/current/alerts-run-osquery.html

### Description

- Outdated information mentioned in the step 5 for Osquery timeout field, As for now it got timed out as per the value we fill in the timeout field.

**Screen
![image (1)](https://github.com/elastic/security-docs/assets/108654988/b0284d14-7d7a-4123-b6bf-344c3e0c9f50)
shot**

### Which documentation set(s) does this bug apply to?

ESS only

### Release version

This bug applies to the stack version 8.12.0

muskangulati-qasource:
Reviewed and assigned to @jmikell821

nastasha-solomon:
@szwarckonrad do you happen to know if the 5-minute timeout period was the default and preset setting for Osquery live queries before 8.12? I checked the Kibana docs, and in versions 8.4-8.11, we tell users that queries will automatically timeout if there have been no responses after 5 minutes. For reference, see the second sentence in step 6 [here](https://www.elastic.co/guide/en/kibana/8.11/osquery.html#osquery-run-query).

nastasha-solomon:
Thanks for the review, @sukhwindersingh-qasource! The docs are ready for review at https://github.com/elastic/security-docs/pull/4658 when you have a chance.

nastasha-solomon:
Docs updated so closing this issue. Will follow up on [this](https://github.com/elastic/security-docs/issues/4626#issuecomment-1906992687) separately.

PR #4658, opened 2024-01-23
[BUG][ESS] Outdated information mentioned in the step 5 for Osquery

nastasha-solomon:
Contributes to https://github.com/elastic/security-docs/issues/4626

[Preview](https://security-docs_bk_4658.docs-preview.app.elstc.co/guide/en/security/master/alerts-run-osquery.html) - Updated step 5

Twin Serverless PR: https://github.com/elastic/staging-serverless-security-docs/pull/274

### DIFF
--- a/docs/osquery/alerts-run-osquery.asciidoc
+++ b/docs/osquery/alerts-run-osquery.asciidoc
@@ -35,7 +35,7 @@ TIP: Refer to {kibana-ref}/osquery.html#osquery-prebuilt-packs-queries[prebuilt 
 [role="screenshot"]
 image::images/setup-query.png[width=80%][height=80%][Shows how to set up a single query]
 
-. Click **Submit**. Query results display within the flyout.
+. Click **Submit**. Query results will display within the flyout.
 +
 NOTE: Refer to <<view-osquery-results>> for more information about query results.
 . Click *Save for later* to save the query for future use (optional).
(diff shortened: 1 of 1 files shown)


docs/osquery/alerts-run-osquery.asciidoc
```diff
@@ -35,7 +35,7 @@ TIP: Refer to {kibana-ref}/osquery.html#osquery-prebuilt-packs-queries[prebuilt
 [role="screenshot"]
 image::images/setup-query.png[width=80%][height=80%][Shows how to set up a single query]
 
-. Click **Submit**. Queries will time out after 5 minutes if there are no responses. Otherwise, query results display within the flyout.
+. Click **Submit**. Query results will display within the flyout.
```
natasha-moore-elastic, line 38: I wonder if it's worth keeping the information about queries timing out. If the default is now 1 minute, we could say something like:

> Query results will display within the flyout. If there are no responses, queries will time out after one minute, unless a custom timeout period was set.

It may not be necessary, though, since we do mention in step 4 that the timeout setting's default value is 60. I'll leave it up to you :)

nastasha-solomon: yeah, I had considered this, but I think mentioning the timeout setting and default configuration in step 4 is sufficient. : )

joepeeples (approved):

PR #4683, opened 2024-01-25
Documents that nested fields aren't supported for Theshold rules "Group by" operations

### LINKED ISSUE #4119: [BUG] Document unsupported field types for threshold rule type `Group by` feature
Related tickets: https://github.com/elastic/kibana/issues/170168, https://github.com/elastic/sdh-security-team/issues/758 (internal)

We currently don't support nested field types for our threshold rule type's Group by feature but don't specify this in the UI or in the [documentation for threshold rules](https://www.elastic.co/guide/en/security/current/rules-ui-create.html#create-threshold-rule). We should add a clarification of this limitation until an enhancement is ever implemented down the line.

benironside:
Resolves #4119 by adding a note that says you can't use nested fields for "group by" to the doc about how to make a threshold rule.

Please let me know if you think we should add any other details. One thing we could consider adding a link to this page: https://www.elastic.co/guide/en/elasticsearch/reference/8.12/nested.html

Preview: Create a threshold rule

### DIFF
--- a/docs/detections/rules-ui-create.asciidoc
+++ b/docs/detections/rules-ui-create.asciidoc
@@ -108,6 +108,9 @@ alerts.
 NOTE: You can use {kib} saved queries (image:images/saved-query-menu.png[Saved query menu,18,18]) and queries from saved Timelines (*Import query from saved Timeline*) as rule conditions.
 
 .. Use the *Group by* and *Threshold* fields to determine which source event field is used as a threshold and the threshold's value.
++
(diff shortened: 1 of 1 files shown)


joepeeples (approved):
LGTM! I'm leaning away from adding a link to the [nested fields](https://www.elastic.co/guide/en/elasticsearch/reference/8.12/nested.html) docs. The concept seems easy to understand on its own, and that page gets real technical real fast, so it feels like extra homework for the reader instead of giving more clarity.

Issue #4653, opened 2024-01-23
[Enhancement]: Add docs for data view on Alerts page


e40pud:
### Description

We need to extend documentation about data view selection on Alerts page. Looks like we would need to add a new section [here](https://www.elastic.co/guide/en/security/8.12/data-views-in-sec.html#_switch_to_another_data_view).

A user asks whether it is possible to update "Security Data View" which is used on "Alerts" page to add remote alerts’ indices. Right now, we show only `.alerts-security.alerts-default` and customer wants to have `.alerts-security.alerts-default, remote-cluster:.alerts-security.alerts-default` index patterns, to see alerts from remote clusters as well.

**UPDATED:**
It is not possible to change data view for the Alerts page and the data view for the page bound to `.alerts-security.alerts-default` index pattern.
Would be nice to add a docs note about that.

![298594598-cd4af9e1-4d9a-4ebe-99af-bee4f81507e9-2](https://github.com/elastic/security-docs/assets/2700761/3835cdb5-61f8-4dd4-8dab-e61c416a9347)

### Which documentation set needs improvement?

ESS and serverless

### Software version

This functionality was introduced in 8.0, so ideally we would add these new docs 8.0+ or any possible version starting from 8.0.

### Collaborators

PM:
Designer: 
Developer:
Others (if applicable):

### Timeline / deliverables

If we can add these docs in 8.13 that would be awesome.

PR #4695, opened 2024-01-26
Documents how to change the default security index

benironside:
First pass at fixing #4653.

@e40pud please let me know what you think, if we should add any more info and if this addresses the need. Thanks!


e40pud:
@benironside sorry, I did not really clarify well what docs we want to add. I updated the original ticket and added **UPDATED** section there:

> It is not possible to change data view for the Alerts page and the data view for the page bound to `.alerts-security.alerts-default` index pattern.
> Would be nice to add a docs note about that.

Let me know if you have questions about it.

e40pud (approved):
Thanks for putting this together!

docs/getting-started/data-views-in-sec.asciidoc
```diff
@@ -18,14 +18,17 @@ image::images/dataview-button-highlighted.png[image highlighting how to open the
...
+
+You can also temporarily modify the active {data-source} from the *{data-source-cap}* menu by clicking *Advanced options*, then adding or removing index patterns.
 
 image::images/dataview-filter-example.gif[video showing how to filter the active data view]
 
 This only allows you to add index patterns that match indices that currently contain data (other index patterns are unavailable). Note that any changes made are saved in the current browser window and won't persist if you open a new tab.
 
-To permanently modify a {data-source}, delete an existing {data-source} or create a new one, you need the required permissions.
-To learn more, refer to {apm-app-ref}/data-views.html[{kib} {data-sources-cap}].
+NOTE: You cannot update the data view for the Alerts page. It always shows data from `.alerts-security.alerts-default`.
```
nastasha-solomon, line 31: @e40pud can we tweak the UI copy so it aligns with this note? The current copy could be interpreted as: you must go elsewhere to update the data view for the Alerts page. 

Screenshot 2024-01-30 at 2 55 42 PM


If it's still possible to change the copy, I'd suggest this:

You can't modify the data view for the Alerts page. 

If that version is too blunt and overly formal, here's a slightly softer version:

This page's data view can't be adjusted.

nastasha-solomon (commented):
Looks great! Thanks, @benironside!

docs/getting-started/data-views-in-sec.asciidoc
e40pud: It is not bound to Alerts page only. Same happens on rule details page. Also, there is another similar message that we show on the timeline. So, general message looks better to me.

https://github.com/elastic/security-docs/assets/2700761/f47b0b78-9939-41e0-9691-c097fb695076

nastasha-solomon: Ah, interesting. Thanks for pointing out these other areas @e40pud -- I wasn't aware of them. Would it be ok if we continued this conversation in https://github.com/elastic/security-docs/issues/4718 so that we're not blocking this doc issue from being closed and merged?

e40pud (approved):

nastasha-solomon (approved):

PR #4752, opened 2024-02-05
removes broken link in README

benironside:
Very minor fix removing a broken link in the repo's README.


joepeeples (commented):
While we're fixing this link, how about also changing the link in step 2 below ("Open an issue using the appropriate template") to point to the issue selection page here, instead of the file browser view: https://github.com/elastic/security-docs/issues/new/choose

joepeeples (approved):

PR #3860, opened 2023-09-05
[docs] Highlight that rule exceptions are case-sensitive

toby-sutor:
Rule exceptions are case-sensitive. If the arguments are not configured in the same format as they appear in the log lines, they will not match the exception.

References:
* https://github.com/elastic/sdh-security-team/issues/714
* https://github.com/elastic/enhancements/issues/15976

PR #4805, opened 2024-02-15
[Redo][8.6-8.13] Highlight that rule exceptions are case-sensitive

nastasha-solomon:
Duplicates changes in https://github.com/elastic/security-docs/pull/3860 to ensure the preview generates. I'll also be filing a separate PR to update the 7.17-8.5 exception docs because they're organized differently on those branches.

Previews:
- [Add exceptions to a rule](https://security-docs_bk_4805.docs-preview.app.elstc.co/guide/en/security/master/add-exceptions.html#detection-rule-exceptions)
- [Add Elastic endpoint exceptions](https://security-docs_bk_4805.docs-preview.app.elstc.co/guide/en/security/master/add-exceptions.html#endpoint-rule-exceptions)


benironside (approved):
I'm not sure that including the same warning twice just a few lines apart is necessary, but your call. LGTM

Issue #4794, opened 2024-02-13
8.12.2 Release notes


nastasha-solomon:
### Description
Generate the 8.12.1 release notes and publish them before or on the release day, which is **March 7, 2024**.

**Milestone dates:**
- **FF:** February 20, 2024 
- **BC:** February 22, 2024
- **Release:** March 7, 2024

### Notes:
- [x] @nastasha-solomon to check with @roxana-gheorghe if any Endpoint PRs need to be release-noted for 8.12.2.
- If you add the PR to the Endpoint feature/bug/known issue, use the hardcoded URL to create the reference link. [Here](https://www.elastic.co/guide/en/security/8.3/release-notes-header-8.1.0.html#bug-fixes-8.1.2) is an example.
- ResponseOps features for cases, connectors, and alerts are always doc'd in the Kibana release notes _unless_ they are Security-specific features. This is also the case for Osquery features and enhancements.

nastasha-solomon:
@jmikell821 - quick FYI that @benironside will be following up on release note PR reviews when I'm out of the office next week.

nastasha-solomon:
No endpoint PRs reported.

nastasha-solomon:
Verified with @roxana-gheorghe that there were no endpoint PRs that needed to be doc'd for this release.

PR #4834, opened 2024-02-18
8.12.2 Release notes

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/4794

Preview:
- [Index page](https://security-docs_bk_4834.docs-preview.app.elstc.co/guide/en/security/master/release-notes.html)
- [Release notes](https://security-docs_bk_4834.docs-preview.app.elstc.co/guide/en/security/master/release-notes-header-8.12.0.html#release-notes-8.12.2)

### DIFF
--- a/docs/release-notes/8.12.asciidoc
+++ b/docs/release-notes/8.12.asciidoc
@@ -11,7 +11,7 @@
 * Fixes long-running queries in Timeline and Events tables within Explore pages ({pull}176838[#176838]).
 * Updates the default {bedrock} connector API URL ({pull}176090[#176090]).
 * Ensures the risk score query only searches through alerts associated with the current user ({pull}175903[#175903]).
-* Fixes a bug that prevented numbers from being used for Osquery pack IDs ({pull}176507[#176507]).
+* Fixes a bug that prevented scheduled query packs from running if a pack's ID was composed of numbers ({pull}176507[#176507]).
 * Fixes bugs that affected the rule details page layout if rule filters were extremely long or had aliases ({pull}176590[#176590]).
 * Fixes a bug that prevented rules from being successfully imported if any rules referenced preconfigured connectors ({pull}176284[#176284]).
 * Fixes a bug that prevented rules from being successfully exported if you exported more than 1000 rules ({pull}175979[#175979]).
--- a/docs/release-notes/8.12.asciidoc
+++ b/docs/release-notes/8.12.asciidoc
@@ -8,15 +8,15 @@
 [discrete]
 [[bug-fixes-8.12.2]]
 ==== Bug fixes
-* Fixes long running queries in Timeline and in Events tables within Explore pages ({pull}176838[#176838]).
+* Fixes long-running queries in Timeline and Events tables within Explore pages ({pull}176838[#176838]).
 * Updates the default {bedrock} connector API URL ({pull}176090[#176090]).
 * Ensures the risk score query only searches through alerts associated with the current user ({pull}175903[#175903]).
 * Fixes a bug that prevented numbers from being used for Osquery pack IDs ({pull}176507[#176507]).
 * Fixes bugs that affected the rule details page layout if rule filters were extremely long or had aliases ({pull}176590[#176590]).
-* Fixes a bug that prevented rules from being imported successfully if the rule referenced a preconfigured connector ({pull}176284[#176284]).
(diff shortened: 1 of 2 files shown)


docs/release-notes/8.12.asciidoc
```diff
@@ -2,15 +2,32 @@
...
+=== 8.12.2
 
-* Provides performance improvements related to image load and registry write events ({pull}175486[#175486]).
+[discrete]
+[[bug-fixes-8.12.2]]
+==== Bug fixes
+* Fixes long-running queries in Timeline and Events tables within Explore pages ({pull}176838[#176838]).
+* Updates the default {bedrock} connector API URL ({pull}176090[#176090]).
+* Ensures the risk score query only searches through alerts associated with the current user ({pull}175903[#175903]).
+* Fixes a bug that prevented numbers from being used for Osquery pack IDs ({pull}176507[#176507]).
```
tomsonpl, line 14: Thanks for taking care of this 🙇 
I think I explained it poorly in my PR, sorry! I would say it can be more like:
`Fixes a bug that prevents scheduled pack's query to run if query id is made of only numbers.` 
Does this make sense?

docs/release-notes/8.12.asciidoc
```diff
@@ -2,15 +2,32 @@
...
+* Fixes bugs that affected the rule details page layout if rule filters were extremely long or had aliases ({pull}176590[#176590]).
```
jpdjere, line 15: Not sure how much detail should go into this but these are actually two different bugs:
`Fixes bug that affected the rule details page layout if rule filters were extremely long`
`Fixes bug that affected rule filters in rule details page if the filters had aliases`

jpdjere (dismissed):
Left a comment, but letting up to you if the detail if needed. 👍

maximpn (dismissed):
LGTM

docs/release-notes/8.12.asciidoc
banderror: ++ These are two different bugs fixed by a single PR. The second one, related to filter aliases, had nothing to do with the page's layout.

Also FWIW "aliases" are called "Custom labels" in the UI.

banderror (approved):
LGTM 👍 
Thank you @nastasha-solomon!

docs/release-notes/8.12.asciidoc
nastasha-solomon: Thanks for the additional context here, @tomsonpl! How's this:
```suggestion
* Fixes a bug that prevented scheduled query packs from running if a pack's ID was composed of numbers ({pull}176507[#176507]).
```

docs/release-notes/8.12.asciidoc
nastasha-solomon: Thanks, both! How about this:

```suggestion
* Fixes a bug that affected the rule details page layout if rule filters were extremely long. Also fixes a bug that incorrectly caused rule filters to display instead of custom labels ({pull}176590[#176590]).
```

docs/release-notes/8.12.asciidoc
tomsonpl: Looks awesome, thank you :)

tomsonpl (approved):

docs/release-notes/8.12.asciidoc
```diff
@@ -2,15 +2,32 @@
...
-[[enhancements-8.12.1]]
-==== Enhancements
+[[release-notes-8.12.2]]
+=== 8.12.2
 
-* Provides performance improvements related to image load and registry write events ({pull}175486[#175486]).
+[discrete]
+[[bug-fixes-8.12.2]]
+==== Bug fixes
+* Fixes long-running queries in Timeline and Events tables within Explore pages ({pull}176838[#176838]).
```
michaelolo24, line 11: 👍🏾 thanks! Succinct and to the point 😄

michaelolo24 (approved):
Thank you! LGTM 🚀

docs/release-notes/8.12.asciidoc
jpdjere: I'd suggest:

"Also fixes a bug that incorrectly caused rule filters to display instead of **their** custom labels"

The logic is: if the rule filter has a custom label, display that label instead of the rule filter itself. The bug was showing the filter always, even for filters that had custom labels.

Hope that helps.

joepeeples (approved):
LGTM! Thanks for preparing this content!

PR #4839, opened 2024-02-20
[Redo][8.2] Highlight that rule exceptions are case-sensitive 

nastasha-solomon:


docs/detections/detections-ui-exceptions.asciidoc
```diff
@@ -123,10 +123,11 @@ the exception prevents the rule from generating alerts when the
 +
 [IMPORTANT]
 ============
+
```
nastasha-solomon, line 126:
```suggestion
```

PR #4876, opened 2024-03-07
Update role definitions for manual CSPM installation

kubasobon:
Related to changes in https://github.com/elastic/cloudbeat/pull/2009

- **update cloudbeat-root with new inline policy**
- **remove SecurityAudit assignment from cloudbeat-root**


kubasobon:
run docs-build

romulets (approved):

kubasobon:
Actually, we will be fine keeping the `SecurityAudit` with no tags. It will use backwards compatible Cloudbeat code.

PR #4883, opened 2024-03-07
Update configure-integration-policy.asciidoc

joe-desimone:
Clarifies the description of our reputation service

### DIFF
--- a/docs/getting-started/configure-integration-policy.asciidoc
+++ b/docs/getting-started/configure-integration-policy.asciidoc
@@ -169,7 +169,7 @@ Malicious behavior protection levels are:
 * **Prevent** (Default): Detects malicious behavior on the host, forces the process to stop,
 and generates an alert.
 
-Select whether you want to use **Reputation service** for additional malware analysis. This service identifies malicious activity and false positives, and enriches alerts using data from various sources, such as VirusTotal and telemetry. For example, reputation service can detect suspicious downloads of binaries with low or malicious reputation.
+Select whether you want to use **Reputation service** for additional protection. Elastic's **Reputation service** leverages our extensive threat intelligence knowledge to make high confidence real-time block decisions. For example, reputation service can detect suspicious downloads of binaries with low or malicious reputation.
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/configure-integration-policy.asciidoc
```diff
@@ -169,7 +169,7 @@ Malicious behavior protection levels are:
 * **Prevent** (Default): Detects malicious behavior on the host, forces the process to stop,
 and generates an alert.
 
-Select whether you want to use **Reputation service** for additional malware analysis. This service identifies malicious activity and false positives, and enriches alerts using data from various sources, such as VirusTotal and telemetry. For example, reputation service can detect suspicious downloads of binaries with low or malicious reputation.
+Select whether you want to use **Reputation service** for additional protection. Elastic's **Reputation service** leverages our extensive threat intelligence knowledge to make high confidence real-time block decisions. For example, reputation service can detect suspicious downloads of binaries with low or malicious reputation.
```
joepeeples, line 172: No need to capitalize or bold the term when not referring directly to the UI. Also, would "prevention decisions" be an OK replacement for "block decisions"? It might be a little more consistent with other language in the docs.
```suggestion
Select whether you want to use **Reputation service** for additional protection. Elastic's reputation service leverages our extensive threat intelligence knowledge to make high confidence real-time prevention decisions. For example, reputation service can detect suspicious downloads of binaries with low or malicious reputation.
```

joepeeples (commented):
Thanks for this revision @joe-desimone! I have an edit to consider; I'll also add the version label to the PR to make sure this gets into 8.13 docs — feel free to add other versions if this needs to go back further.

docs/getting-started/configure-integration-policy.asciidoc
joe-desimone: both work for me!

joe-desimone:
thanks for the assist @joepeeples !

joepeeples (approved):

PR #4915, opened 2024-03-14
Update supported k8s version for EKS

uri-weisman:
We encountered an [SDH](https://github.com/elastic/sdh-beats/issues/4473) ticket where a customer upgraded their EKS cluster to v1.25 and lost all their Pod Security Policy (PSP) findings because PSPs were removed in this version.
I don't believe we can claim to support all available EKS versions because our benchmark is not being updated, and we are likely to miss relevant findings.
It seems logical to tie our EKS support to the Kubernetes version we officially support. What do you think, @tinnytintin10?

Preview: [KSPM FAQ](https://security-docs_bk_4915.docs-preview.app.elstc.co/guide/en/security/master/security-posture-faq.html)


smriti0321:
@uri-weisman Thanks for starting this conversation, I agree to call out the specific versions we support in our docs, to help get this update could you please highlight the exact Kubernetes versions for our EKS support? cc @benironside

uri-weisman:
@smriti0321 

> could you please highlight the exact Kubernetes versions for our EKS support?


This problem arises from the CIS EKS benchmark v1.0.1, which does not specify the EKS versions to which the benchmark is applicable. For example, in the mentioned benchmark version, they refer to k8s resources (PSP) that are already being removed in the current version (v1.25).

The EKS benchmark implementation includes data collectors that are specific to the cloud provider and some that are shared with self-managed k8s, supporting v1.23.
Therefore, I suggest that we align our support with this version.

This means that we will publish that our supported EKS version is v1.23.The newer versions may work, but they might result in lost findings.

cc @tehilashn

joepeeples (approved):
🏝️

PR #4976, opened 2024-03-22
Switch to global attributes and remove section-scoped `pull` and `issue`

joepeeples:
Changes all instances of book/section-scoped attributes `{pull}` and `{issue}` to global attributes `{kibana-pull}` and `{kibana-issue}`, within the main branch.

Also removes the book/section-scoped definitions from the index file `release-notes.asciidoc`.

> [!NOTE]  
> We also made this change to the 7.17 branch with https://github.com/elastic/security-docs/pull/4978. We didn't update any other currently available branches because they're unlikely to be part of the migration. When we migrate, we'll only be migrating `main`, `7.17`, and whatever the current branch is at that time (probably 8.15 by that point, definitely not 8.13).


jmikell821 (dismissed):
Reviewed this twice; LGTM :)

nastasha-solomon (approved):

PR #5339, opened 2024-06-06
[8.13] Removes all attack discovery pages

### LINKED ISSUE #5335: [Request] remove Attack Discovery 8.13 ESS doc
### Description

The 8.13 documentation is no longer needed, since it was introduced as a stop gap to enable doclinks in the serverless version of the product to work. 

### Background & resources

* PRs: 
* Issues/metas: 
* Point of contact: 
* Test environments: 

### Which documentation set does this change impact?

ESS only

### ESS release

8.13

### Serverless release

-

### Feature differences

-

### API docs impact

-

### Prerequisites, privileges, feature flags

_No response_

benironside:
Fixes #5335 

Attack Discovery docs were added to 8.13 ESS docs even though the feature came out in 8.14. This was to enable the UI link to the docs in the serverless version of the product to work. The 8.13. ESS doc is no longer needed

Note that this will only be backported to 8.13. I'm scratching my head a bit about the what's new page. I think the changes I added should work, but please give that page some thought during your review.

Preview: [LLM performance matrix](https://security-docs_bk_5339.docs-preview.app.elstc.co/guide/en/security/master/llm-performance-matrix.html)


joepeeples:
@benironside I think this PR needs to be adjusted so it targets the `8.13` branch, not `main`. You want to remove the Attack discovery content from the 8.13 docs, but keep them in 8.14 and forward, right?

You can change the target by clicking Edit and then selecting a new branch in the drop down that appears. Sometimes switching targets like this can throw off the PR's diff, but I'm not sure exactly what'll happen. 👀 

image

benironside:
> @benironside I think this PR needs to be adjusted so it targets the `8.13` branch, not `main`. You want to remove the Attack discovery content from the 8.13 docs, but keep it in 8.14 and forward, right?
> 
> You can change the target by clicking Edit and then selecting a new branch in the drop down that appears. Sometimes switching targets like this can throw off the PR's diff, but I'm not sure exactly what'll happen. 👀
> image
> 
> This probably also explains what you said above about What's New, because you're currently using the version of that page from `main` instead of `8.13`. [What's New in 8.13](https://www.elastic.co/guide/en/security/8.13/whats-new.html) didn't have any Gen AI features to begin with, so there wouldn't have been anything you'd need to remove.

That makes sense! Thanks Joe!

benironside:
closing this and reopening it against the 8.13 branch

Issue #4796, opened 2024-02-13
API key authentication for cross-cluster search & detection rules


joepeeples:
### Description

Currently [our cross-cluster search (CCS) docs](https://www.elastic.co/guide/en/security/current/rules-cross-cluster-search.html) mention the API key authentication method but don't explain it in more detail. We can expand the docs to provide more guidelines as a fully supported option.

@vgomez-el has done testing to confirm the method, with these additional caveats:
- Pay attention to the API key expiration. Once the key expires, users won't be able to connect to remote clusters.
- Pay attention to the indices permissions when creating the API key, otherwise users will need to create a new API KEY and start the process again.
- For cloud deployments, all deployments must be on the same region.
- Carefully follow the prerequisites and other [documentation](https://www.elastic.co/guide/en/cloud/current/remote-clusters-api-key.html) for API key auth in CCS.

Confirmed in 8.10, version when API auth was added in general, but for cross-cluster search it was in beta until 8.14. Either backport to 8.14, or backport further to 8.10 and add the beta tag.

### Background & resources

* PRs: n/a
* Issues/metas: ??
* Point of contact: @vgomez-el 
* Test environments: ??

### Which documentation set does this change impact?

ESS only

### ESS release

N/A

### Serverless release

n/a

### Feature differences

n/a

### API docs impact

??

### Prerequisites, privileges, feature flags

??

joepeeples:
New info: API auth for cross-cluster search was added in 8.10, but it was in beta until 8.14. Updated description above with backporting options to consider.

PR #5447, opened 2024-06-18
API key authentication for cross-cluster search & detection rules

joepeeples:
Resolves https://github.com/elastic/security-docs/issues/4796.

No twin PR needed for serverless (not supported yet).

### Preview
- [Cross-cluster search and detection rules](https://security-docs_bk_5447.docs-preview.app.elstc.co/guide/en/security/master/rules-cross-cluster-search.html) — Updates step 1 to include both auth methods, and also adds some disambiguation in the "Update a rule's API key" section.

### Backports
API auth for cross-cluster search was added in 8.10, but it was in **beta** until 8.14. So we can backport this straight to 8.14, but we should add a beta tag and revise slightly in 8.10–8.13.

We should be able to get Mergify to open the backport PRs for all versions, then manually update the ones where we need the beta tag and revised verbiage. 🤞


yctercero (approved):
LGTM - thanks so much and so sorry it took me so long to review.

nastasha-solomon (approved):

PR #5501, opened 2024-07-03
Update .backportrc.json: remove 7.x preset

joepeeples:
Removes syntax that creates a pre-selected (or "checked")`7.x` option every time you run the backport tool. We used to do this during the 7.x series, but we no longer maintain such a branch and don't need it as a pre-selected option.

Backport tool docs: https://github.com/sorenlouv/backport/blob/main/docs/config-file-options.md#targetbranchchoices-required
(The documented setting names are slightly different from what's in our config file. Maybe they were renamed at some point but maintain backwards compatibility with our names?)


jmikell821 (approved):

joepeeples:
FYI: I don't think we need to backport this change to previous branches, since we usually run the backport tool from `main`, not a versioned branch.

PR #5513, opened 2024-07-05
(Doc+) Alerts UI cannot be CCS

stefnestor:
👋 howdy, team! 

Playing forward a couple ballpark Githubs (kindly see their sub-internal-links) https://github.com/elastic/security-docs/issues/4653 , https://github.com/elastic/kibana/issues/172534 , can we expand [this doc's](https://www.elastic.co/guide/en/security/master/data-views-in-sec.html#_create_or_modify_a_data_view) note to more heavily indicate even CCS is not possible on the Alerts UI?


docs/getting-started/data-views-in-sec.asciidoc
```diff
@@ -28,7 +28,7 @@ image::images/dataview-filter-example.gif[video showing how to filter the active
 
 This only allows you to add index patterns that match indices that currently contain data (other index patterns are unavailable). Note that any changes made are saved in the current browser window and won't persist if you open a new tab.
 
-NOTE: You cannot update the data view for the Alerts page. It always shows data from `.alerts-security.alerts-default`.
+NOTE: You cannot update the data view for the Alerts page, e.g. to enable it to reference a cross-cluster search (CCS) data view. It always shows data from `.alerts-security.alerts-default`.
```
joepeeples, line 31: I think the CCS use case would be better as a separate sentence, so the first statement has more impact. Something like this:

```suggestion
NOTE: You cannot update the data view for the Alerts page. This includes referencing a cross-cluster search (CCS) data view or any other data view. The Alerts page always shows data from `.alerts-security.alerts-default`.
```

joepeeples (commented):
Added a suggested edit, LMK what you think or feel free to commit if it looks OK to you.

We should also update the serverless docs with the same clarification; I can push a commit to this PR that'll do that shortly.

joepeeples:
> We should also update the serverless docs with the same clarification; I can push a commit to this PR that'll do that shortly.

@stefnestor As I dug into this more, I noticed some other quirks (mostly unrelated) that I also need to fix, so I'll open a separate PR to update serverless docs _and_ do the other fixes in one swoop. In the meantime, I'll approve this PR and can take care of merging & backporting to keep things moving.

joepeeples (approved):
👍

PR #3054, opened 2023-03-19
Update docs to show that CCS Supported for IM rules in 8.7

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/3044.

Previews:
- Limited support for indicator match rules
- Compatibility with cold tier nodes

Issue #551, opened 2021-03-15
[DOCS] Document to set limitations on what we support for Indicator Match rules in 7.12


dontcallmesherryli:
## Description 

Don't use cold storage, make sure your indicator index doesn't have X, don't use cross cluster search
mostly elastic search; item search 9000
SDH troubleshooting guide

## Acceptance Test Criteria

List all the ATC of each action and its intended result. 
As a user, when [action (e.g., viewing, clicking, selecting, etc.)] the [insert the expected result]. 
If the doc issue includes a procedure, number the steps in sequential order.  

## Notes

- Add the **"Team:Docs"** label to new issues. 
- Be sure to add the version number label. 
- Be sure to add any necessary screenshots for clarity. 
- Include any conditions or caveats that may affect customers.

jmikell821:
@dontcallmesherryli can I have some more context around what needs to be documented? [Here](https://www.elastic.co/guide/en/security/7.11/detection-engine-overview.html#cold-tier-detections) is what we say about cold tier storage. [Here](https://www.elastic.co/guide/en/security/7.11/rules-ui-create.html#create-indicator-rule) is our topic on indicator match rules.

dontcallmesherryli:
Hi @jmikell821 I'm working on this ticket still adding details. That's why I didn't tag this ticket with any labels or assignees yet. Will tag you when it's ready! Thank you

MikePaquette:
Note: Cold Tier Storage documentation change requests are listed in this Issue https://github.com/elastic/security-docs/issues/562 

This ticket includes changes to **four** sections related to Indicator Match rules:

- [ ] 1. On the [Detections and Alerts page](https://www.elastic.co/guide/en/security/7.12/detection-engine-overview.html) - Add a new section (probably after the compatibility with cold tier data section and before the Terminology section)  called "Limited Support for Indicator Match Rules".  The section would have the following points:

Indicator Match rules provide a powerful capability to search your security data, however, their queries can consume significant deployment resources.  Therefore the following support restrictions are in place:
* Elastic Security does not support the use of Cold Tier Data with Indicator Match rules.
* The use of Cross Cluster Search (CCS) with Indicator Match rules is not supported.
* Indicator Match rules with an Additional Look-back time value of greater than 24 hours are not supported.
* ~~Indicator indices with more than 90,000 items are not supported.~~

[...]

PR #5564, opened 2024-07-15
(Doc+) SIEM + frozen tier compatibility 

stefnestor:
👋🏽 howdy, team! 

1. Since [ES Data Tiers](https://www.elastic.co/guide/en/elasticsearch/reference/current/data-tiers.html) are sequentially hardware performant, because we call out `cold` this appends the same rules apply to `frozen` (related to `securitySolution:excludeColdAndFrozenTiersInAnalyzer` from [kibana#172162](https://github.com/elastic/kibana/pull/172162)).  Kindly see examples 
    - [kibana#139969](https://github.com/elastic/kibana/issues/139969) for linked internal examples of e.g. alert indices `.alerts-security*` being hosted in frozen breaking SIEM. 
    - [kb#39606517](https://support.elastic.co/knowledge/39606517) where frozen indices containing future dates catch in the Rule lookback window and therefore can degrade product performance. 
2. It looks like there's a duplicate section in the [SIEM](https://www.elastic.co/guide/en/security/master/detection-engine-overview.html#cold-tier-detections) and [Kibana](https://www.elastic.co/guide/en/kibana/current/xpack-siem.html#_compatibility_with_cold_tier_nodes) docs from [security-docs#487](https://github.com/elastic/security-docs/pull/487) and [kibana#103151](https://github.com/elastic/kibana/pull/103151), respectively, where the latter had additional phrasing so cross-pollinated. 
    - ( Paper-trail FYI for my team: [security-docs#551](https://github.com/elastic/security-docs/issues/551) later appears to confirm indicator rules don't support cold, to confirm between PR's which is latest state in (2). It does also suggest indicator match rules do not support cross cluster search, but that was purposely later removed in [security-docs#3054](https://github.com/elastic/security-docs/pull/3054) for v8.7 . )

Preview:
- [Detections and alerts](https://security-docs_bk_5564.docs-preview.app.elstc.co/guide/en/security/master/detection-engine-overview.html)


docs/detections/detection-engine-intro.asciidoc
```diff
@@ -62,20 +62,21 @@ To make sure you can access Detections and manage rules, see
...
 * Index patterns specified in `securitySolution:defaultIndex`
-* Index patterns specified in the definitions of detection rules
+* Index patterns specified in the definitions of detection rules, except for indicator match rules 
 * Index patterns specified in the data sources selector on various {security-app} pages
 
-{elastic-sec} does *NOT* support cold tier data for the following {es} indices:
+{elastic-sec} does *NOT* support either cold or frozen tier data for the following {es} indices:
 
 * Index patterns controlled by {elastic-sec}, including alerts and list indices
+* Index patterns specified in indicator match rules
```
nastasha-solomon, line 77: Adds a bit more consistency in the way we talk about rules on this page.

```suggestion
* Index patterns specified in the definition of indicator match rules 
```

nastasha-solomon (commented):
Thanks for making these changes, @stefnestor! Couple of questions, in addition to the comments that I left:
- Do these changes need to be reflected in the [Serverless version](https://www.elastic.co/docs/current/serverless/security/detection-engine-overview) of this page too? If so, I'm happy to add those updates to this PR. 
- How far back should we port these changes? For example, is this information relevant to `8.7` - `8.15`? 

Thanks again for your help with keeping the Security docs current and correct!

docs/detections/detection-engine-intro.asciidoc
```diff
@@ -62,20 +62,21 @@ To make sure you can access Detections and manage rules, see
 ==============
 
 [float]
-[[cold-tier-detections]]
-== Compatibility with cold tier nodes
+[[data-tier-detections]]
+== Compatibility with cold and frozen tier nodes
 
-Cold tier is a {ref}/data-tiers.html[data tier] that holds time series data that is accessed only occasionally. In {stack} version >=7.11.0, {elastic-sec} supports cold tier data for the following {es} indices:
+Cold and frozen tier are {ref}/data-tiers.html[data tier] that holds time series data that are accessed only occasionally and rarely, respectively. In {stack} version >=7.11.0, {elastic-sec} supports cold tier data for the following {es} indices:
```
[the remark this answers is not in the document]
stefnestor: Agreed not sure we need the full caveat but FWIW, the ES docs draw the distinction:

> [Data Tiers](https://www.elastic.co/guide/en/elasticsearch/reference/current/data-tiers.html) 
> image

stefnestor:
Okay, read again! 🙏

docs/detections/detection-engine-intro.asciidoc
```diff
@@ -63,19 +63,20 @@ To make sure you can access Detections and manage rules, see
 
 [float]
 [[cold-tier-detections]]
-== Compatibility with cold tier nodes
+== Compatibility with cold and frozen tier nodes
 
-Cold tier is a {ref}/data-tiers.html[data tier] that holds time series data that is accessed only occasionally. In {stack} version >=7.11.0, {elastic-sec} supports cold tier data for the following {es} indices:
+Cold and frozen {ref}/data-tiers.html[data tiers] hold time series data that are only accessed occasionally. In {stack} version >=7.11.0, {elastic-sec} supports cold tier data for the following {es} indices:
```
nastasha-solomon, line 68: Oops, forgot to fix this earlier.

```suggestion
Cold and frozen {ref}/data-tiers.html[data tiers] hold time series data that is only accessed occasionally. In {stack} version >=7.11.0, {elastic-sec} supports cold tier data for the following {es} indices:
```

docs/detections/detection-engine-intro.asciidoc
```diff
@@ -63,19 +63,20 @@ To make sure you can access Detections and manage rules, see
 
 [float]
 [[cold-tier-detections]]
-== Compatibility with cold tier nodes
+== Compatibility with cold and frozen tier nodes
 
-Cold tier is a {ref}/data-tiers.html[data tier] that holds time series data that is accessed only occasionally. In {stack} version >=7.11.0, {elastic-sec} supports cold tier data for the following {es} indices:
+Cold and frozen {ref}/data-tiers.html[data tiers] hold time series data that is only accessed occasionally. In {stack} version >=7.11.0, {elastic-sec} supports cold tier data for the following {es} indices:
```
rylnd, line 68: I agree with adding mention of frozen tiers for clarity, but it doesn't seem like this line things much clearer. If I'm understanding correctly, we say:

1. X supports cold tier
2. Y and Z do not support cold or frozen tier

It seems like we should be explicit on whether X (which is the list on lines 70-72) supports frozen tier or not.

stefnestor:
```suggestion
Cold and frozen {ref}/data-tiers.html[data tiers] hold time series data that is only accessed occasionally. In {stack} version >=7.11.0, {elastic-sec} supports cold but not frozen tier data for the following {es} indices:
```

Does this look good?

rylnd: It addresses my issue,  but I defer to @marshallmain to confirm that this statement is correct

marshallmain: This is roughly correct but I think we on the dev/product side need to come up with a more precise concept of what exactly we support because separating things by index pattern isn't sufficient granularity IMO. We do support frozen tier data existing in index patterns queried by detection rules as long as the frozen data is not in the time range queried by the rule. In general for the use cases listed here it's important that frozen tier data isn't queried regularly, i.e. that the time range on the query excludes frozen tier data (or a _tier filter? I can't remember if we tested to confirm that optimization works or not), but cold/frozen tier data is allowed to live in those index patterns.

So for detection rules, we might say "frozen tier data must not be in the time range queried by the rule or it must be explicitly excluded with a _tier filter". For new terms rules, we should also specify that frozen tier data can't be in the (interval + lookback) OR the history window. Similarly for the "data sources selector on various pages", the frozen data only needs to be excluded from the query.

marshallmain: I don't think we need to hold up this PR to rework the concepts here though.

nastasha-solomon (approved):

docs/detections/detection-engine-intro.asciidoc
nastasha-solomon: @marshallmain I'm tracking your feedback in a private docs issue so we don't forget to revisit these docs in the future.

https://github.com/elastic/security-docs-internal/issues/27

PR #5597, opened 2024-07-23
Revises IA of the getting started section 

benironside:
This PR fixes issue 23 in our docs projects repo. Updates the ESS docs Getting Started section for 8.14 and 8.15. 

- Moves some requirements pages to the sections about their corresponding features (not all reqs pages were moved, I didn't make the new section for Defend yet)
- Moves the Elastic Security UI page into the Getting Started section
- Other minor revisions to the Getting Started

Note: This PR affects organization only. It does not make content updates.

Previews: [Getting started section](https://security-docs_bk_5597.docs-preview.app.elstc.co/guide/en/security/master/getting-started.html)


docs/cases/cases-index.asciidoc
```diff
@@ -1,5 +1,7 @@
 include::cases-overview.asciidoc[leveloffset=+1]
 
+include::cases-req.asciidoc[leveloffset=+1]
```
nastasha-solomon, line 3: I think it makes more sense to place it under Cases, and to have it on the same level as the two other pages (Open and manage cases and Configure case settings):

```suggestion
include::cases-req.asciidoc[leveloffset=+2]
```

docs/detections/detections-index.asciidoc
```diff
@@ -2,6 +2,8 @@ include::detection-engine-intro.asciidoc[]
 
 include::about-rules.asciidoc[]
 
+include::detections-req.asciidoc[leveloffset=+1]
```
nastasha-solomon, line 5: I'm debating whether we should swap this page with the "About detection rules" page. The "Detections prerequisites and requirements" is about privs needed to use rules _and alerts_, so it might be more appropriate to place it closer to the "Detections and alerts" page, instead of in between rule-related content.

nastasha-solomon (commented):
Thank you for re-orging this! It looks great 👍🏼  Just left two suggestions for your consideration.

docs/detections/detections-index.asciidoc
benironside: I'm open to that! Other reviewers?

docs/advanced-entity-analytics/advanced-entity-analytics-overview.asciidoc
```diff
@@ -8,11 +8,13 @@ Advanced Entity Analytics provides two key capabilities:
 * <<entity-risk-scoring, Entity risk scoring>> 
 * <<advanced-behavioral-detections, Advanced behavioral detections>>
 
+include::ers-req.asciidoc[leveloffset=+1]
 include::entity-risk-scoring.asciidoc[leveloffset=+1]
 include::asset-criticality.asciidoc[leveloffset=+2]
```
natasha-moore-elastic, line 13: I'd put the entity risk scoring reqs inside the Entity risk scoring section, since it only applies to that part of Entity Analytics
```suggestion
include::entity-risk-scoring.asciidoc[leveloffset=+1]
include::ers-req.asciidoc[leveloffset=+2]
include::asset-criticality.asciidoc[leveloffset=+2]
```

docs/detections/detections-index.asciidoc
natasha-moore-elastic: I agree, that makes sense!

natasha-moore-elastic (commented):
One suggestion about EA, otherwise looks great! 🚀

docs/detections/detections-index.asciidoc
joepeeples: Agree!

jmikell821: I agree with this.

joepeeples (approved):
LGTM! Though I was expecting a bit more of an overhaul, esp compared to the outlines in the originating issue and the Google doc, with new "Get started" subsections for each product area (SIEM, Endpoint, Cloud). I assume that'll come in later iterations, and this PR is just the start of the reorg.

Also, maybe a bit out of scope, but we should standardize whether we call things "requirements" or "prerequisites" (or [both](https://security-docs_bk_5597.docs-preview.app.elstc.co/guide/en/security/master/detections-permissions-section.html)!), in topic titles and anywhere else. The difference between terms is pretty nuanced and probably lost on most users, so we should just pick one and stick with it. Nothing that needs to go into this PR; we can do that later when we're focusing more on content and not just organization.

docs/getting-started/index.asciidoc
```diff
@@ -7,31 +7,38 @@ Looking to get started with {elastic-sec}? This section describes the {elastic-s
...
+
 include::endgame-sensor-reqs.asciidoc[leveloffset=+1]
 include::endgame-sensor-FDA.asciidoc[leveloffset=+2]
 include::endgame-sensor-FDA-ven.asciidoc[leveloffset=+2]
+
 include::security-spaces.asciidoc[leveloffset=+1]
+
 include::data-views-in-sec.asciidoc[leveloffset=+1]
+
 include::ingest-data.asciidoc[leveloffset=+1]
```
jmikell821, line 22: I would put ingest after `Elastic Endgame reqs`, closer to the top. Here's the proposed order according to our [doc](https://docs.google.com/document/d/1YlhTZRWjx__Y2-5R-ndiq9mgc_ZejGD3RcHbXzrfYoQ/edit?usp=sharing) but I'm open to suggestions.

jmikell821 (commented):
Thanks! Left a comment. Also - does this PR move the feature-specific reqs to their respective sections? Or are we doing that in phase 1B?

docs/getting-started/index.asciidoc
benironside: "does this PR move the feature-specific reqs to their respective sections?" - Yes, the only exception being the Elastic Defend stuff since we haven't made a section for it yet.

nastasha-solomon (approved):
lgtm - thank you for reorging this!

Issue #5624, opened 2024-07-30
[Request][8.15 & Serverless] xMatters connector available for rule actions


nastasha-solomon:
### Description

Users can now set up xMatters connector to send actionable alerts to on-call xMatters resources.

### Background & resources

* PRs: https://github.com/elastic/kibana/pull/172933
* Issues/metas: https://github.com/elastic/kibana/issues/124764
* Point of contact: @vitaliidm 
* Test environments: N/A

### Which documentation set does this change impact?

ESS and serverless

### ESS release

8.15

### Serverless release

N/A

### Feature differences

N/A

### API docs impact

TBD

### Prerequisites, privileges, feature flags

N/A

#### Serverless
- [x] Update screenshots in the sections for setting alert notifications and response actions for rules. These changes will be applied when https://github.com/elastic/security-docs/pull/5615 is merged.

#### 8.15

- [x] Update screenshots in the sections for setting alert notifications and response actions for rules.

PR #5625, opened 2024-07-30
[8.15] xMatters connector available for rule actions

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/5624

Previews:
- [Create a detection rule](https://security-docs_bk_5625.docs-preview.app.elstc.co/guide/en/security/master/rules-ui-create.html): Refreshed images that showed available rule action connectors


vitaliidm (approved):
We also add Server log in connectors.
But I don't see mentioning of xMatters anywhere in text, so should be fine for Server Log too

nastasha-solomon:
Yeah, specific connectors aren't mentioned in the UI docs. Instead, the docs link users to the Kibana connector [docs](https://security-docs_bk_5625.docs-preview.app.elstc.co/guide/en/kibana/master/action-types.html) for more info on connectors that are available for rule actions.

joepeeples (approved):
LGTM, but I'd also challenge the idea that we need these screenshots at all. We end up having to update them all the time, and IMO they don't add a ton of value. They illustrate something that the user can very easily see themselves.

nastasha-solomon:
@joepeeples yeah, I agree. I was actually thinking of removing the screenshot in ESS and Serverless docs when I update the 8.16 docs for case system actions. There are a couple of other minor changes that I need to make to the section anyways, so that'll be a good time to do some light cleanup.

Issue #5711, opened 2024-08-18
[Docs IA] Create Investigation Tools section


benironside:
### Description

As part of the Security Docs IA [meta issue](https://github.com/elastic/docs-projects/issues/236), update the "Investigations" section of ESS docs to "Investigation tools", and change its contents.

The goal-state for this section is described in the [IA outline](https://docs.google.com/document/d/1TJ-0W1agfywm22dVD4JqID3NHYwxny-s9U3kafqle3A/edit), though this goal-state may be altered at Nastasha's discretion as she works on it. The [on-week tracker ](https://docs.google.com/spreadsheets/d/1r9_7Cm8f9K0YI5CTEwR8dpxryVzbopYzYOnk0Pht0bk/edit?gid=0#gid=0)also has an analysis of each page. The focus of this issue is reorganization, not content updates, however if you notice any "low hanging fruit" for context setting/introduction in any of the pages, feel free to take this opportunity to edit them.

Changes should be backported to 8.14 and 8.15.

### Background & resources

* PRs: 
* Issues/metas: 
* Point of contact: 
* Test environments: 

### Which documentation set does this change impact?

ESS only

### ESS release

8.14

### Serverless release

-

### Feature differences

-

### API docs impact

-

PR #5736, opened 2024-08-22
[Docs IA][ESS] Create Investigation Tools section 

nastasha-solomon:
Fixes https://github.com/elastic/security-docs/issues/5711 and makes a few notable changes:
- Renames the "Investigate" section to "Investigation tools" and "About Timeline templates" to "Timeline templates" 
- Moves the following topics into the "Investigations tools" section:
   - Visual event analyzer
   - Session view
   - Osquery 
   - Indicators of compromise
- Moves the Cases topic to the end of the "Investigation tools" section so the other threat hunting/event investigation tools are grouped together

Preview: [Investigation tools](https://security-docs_bk_5736.docs-preview.app.elstc.co/guide/en/security/master/investigations-tools.html)


docs/events/timeline-templates.asciidoc
```diff
@@ -1,5 +1,5 @@
 [[timeline-templates-ui]]
-= About Timeline templates
+= Use Timeline templates
```
joepeeples, line 2: The verb is definitely more active than "About," but "Use" is still pretty generic and doesn't add a lot (most topics are about how to "use" some feature). Maybe keep it simple with no verb?
```suggestion
= Timeline templates
```

docs/cloud-native-security/session-view.asciidoc
```diff
@@ -15,7 +15,7 @@ and investigating session activity on your Linux infrastructure and understandin
 
 [float]
 [[session-view-data]]
-====== Session View displays:
+== Session View displays
```
joepeeples, line 18: This is sort of an odd heading because it's also acting as a lead-in sentence to the list below, as in "Session View displays the following things." Because it's unexpected for a heading to also act like a sentence, I was reading it more like "This is a section about the Session View displays" but I don't think that's a thing. Anyway, maybe something like:
```suggestion
== Session View features
```

And/or give the list a proper lead-in sentence, but that feels like adding a sentence just to add a sentence so maybe just focus on the heading.

joepeeples (commented):
Great reorg! Added two quibbles about headings but I really like the new organization.

I'm also a little curious about the specific order of the main subsections within "Investigation Tools": it feels a little random to me but I can't think of any stronger alternatives. Maybe that's it's just kind of a catch-all category of assorted tools.

docs/events/timeline-templates.asciidoc
nastasha-solomon: I'm fine with either, really. Outside of the reasons that I listed in the PR description, I made this change to align this with the pattern used for the Cases and Osquery sections (top-level topic title is the feature name and the sub-topics are use-case-oriented).

nastasha-solomon:
> I'm also a little curious about the specific order of the main subsections within "Investigation Tools": it feels a little random to me but I can't think of any stronger alternatives. Maybe that's it's just kind of a catch-all category of assorted tools.

@joepeeples the tools are loosely ordered from generic to specific: 
- I put Timeline first because it can be used to investigate events from any data source and provides a variety of ways to query and visualize ingested data. 
- Visual event analyzer and Session View are somewhat grouped together because they allow you to visually examine process relationships and user/host behavior. I only put Visual event analyzer first because it relies on the Elastic Defend integration, and I thought the higher placement promoted it a bit more.
- Osquery and Indicators of Compromise don't fall into either of the aforementioned categories so I put them last. They're also tools that give you raw data that you then have to pivot elsewhere to use. For example, Osquery can retrieve information on processes running in your environment, but it doesn't offer much functionality for manipulating that data. If you want to visualize or interact with the data, you have to pivot to Timeline, Discover, or Lens.

joepeeples:
@nastasha-solomon Thanks for the explanation, it all makes sense to me!

> I only put Visual event analyzer first because it relies on the Elastic Defend integration

☝️ TIL this — I've never worked with the event analyzer much or touched its docs. Maybe we could cross-promote analyzer from the Defend docs more, and vice versa, but nothing we need to do for this initial reorg PR. Gonna slap a ✅ on this one.

joepeeples (dismissed):

natasha-moore-elastic (dismissed):
Revamp looks great! The only suggestion, as a potential future improvement, would be to consider whether there is enough overlap between the Timeline and Timeline template pages to combine them into a single page.

nastasha-solomon:
Good suggestion, @natasha-moore-elastic! I doc'd it [here](https://docs.google.com/spreadsheets/d/1r9_7Cm8f9K0YI5CTEwR8dpxryVzbopYzYOnk0Pht0bk/edit?gid=0#gid=0&range=E130) so we don't forget about it as we continue our IA revision efforts.

docs/cloud-native-security/session-view.asciidoc
jmikell821: I would prefer to leave this as `Session View` to align with the other titles of investigation tools -- `Timeline`, `Visual event analyzer`, etc. It would seem a little misaligned if we did not, IMO.

nastasha-solomon: @jmikell821 to clarify, are you suggesting that [this section](https://security-docs_bk_5736.docs-preview.app.elstc.co/guide/en/security/master/session-view.html#session-view-data) be renamed from "Session view features" to "Session View"?

jmikell821 (dismissed):
This looks GREAT! I only added one comment to an ongoing suggestion.

docs/cloud-native-security/session-view.asciidoc
jmikell821: @nastasha-solomon oops — so sorry, my eyes crossed here. I think we can remove this subheading and have a lead-in sentence instead.

nastasha-solomon: All good and thanks, @jmikell821! I've made your requested changes in [284f14b](https://github.com/elastic/security-docs/pull/5736/commits/284f14b651bb4d093f30ff0beaaa1a96a7436d5f).

joepeeples (approved):
🚀 🚀 🚀

PR #5737, opened 2024-08-23
Update response-actions.asciidoc

tammytorbert:
Added full string that needs to be added to kibana.yml to adjust file upload size limit.  Original was missing xpack.securitySolution

### DIFF
--- a/docs/management/admin/response-actions.asciidoc
+++ b/docs/management/admin/response-actions.asciidoc
@@ -166,7 +166,7 @@ Example: `upload --file --comment "Upload remediation script"`
 
 TIP: You can follow this with the `execute` response action to upload and run scripts for mitigation or other purposes.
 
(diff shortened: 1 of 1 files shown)


joepeeples:
Thanks for the update, @tammytorbert, LGTM! I'll take it from here — I'll add another commit to update the same content in our [serverless docs](https://www.elastic.co/docs/current/serverless/security/response-actions), and I'll handle merging and backporting, too.

joepeeples:
Mostly to self: Interestingly, I can't update serverless from this PR, because it targets `8.15` as the base branch, and we don't keep serverless files in versioned branches. (This triggered the check-and-delete-serverless workflow GH Action.)

So, I'll go ahead and merge this to `8.15`, and open a separate PR targeting `main` to update the content in serverless. For backports, I could either let Mergify spin them up from my `main` PR, or use the backport tool to spin up from the `8.15` PR. Probably the former, to give the robots something to keep busy with.

joepeeples (approved):

Issue #5097, opened 2024-04-17
[Enhancement]: Document the endpoint "top" command


nfritts:
### Description

We've added command line context/help for the `top` command in 8.12 but think the documentation should also be in the official docs.

### Related links / assets

Please include each of the following, if applicable:
Doc URL: 
Subject matter expert:
Figma link(s):
Github epic link(s):
Github issue link(s): https://github.com/elastic/endpoint-dev/issues/13771

### Which documentation set needs improvement?

ESS and serverless

### Software version

8.12+

### Collaborators

PM:
Designer: 
Developer:
Others (if applicable):

### Timeline / deliverables

No specific timeline really.

intxgo:
Below is output of 8.15.0 `help` command, and my draft of the documentation page. I kept the format of Elastic Agent command reference.

Feel free to ask for any clarification. The documentation needs to be slightly updated for backports if we want to do it as we've been changing a bit the commands and their output over time.

=============================

```
C:\Users\le>"\Program Files\Elastic\Endpoint\elastic-endpoint.exe" help
Elastic Endpoint 8.15.0-SNAPSHOT
www.elastic.co

Usage: elastic-endpoint.exe command [sub-command] [common options]

commands:
    version
        description: print version details

    install
        --resources          install resources zip
        --upgrade            upgrade existing installation
        description: install Elastic Endpoint

    uninstall
        --uninstall-token    tamper protection token
        description: uninstall Elastic Endpoint

    run
        description: run elastic-endpoint.exe as foreground process
                     if no other instance is already running

    memorydump
        --compress           compress saved memory dump
        --timeout            default 60 seconds
        description: collect Endpoint's service memory dump

    send
        metadata             send off-schedule metrics
        description: send requested document to the stack

    test
        output               test if Endpoint can connect to remote resources
        description: perform requested test

    diagnostics
        description: collect diagnostics

    top
        --interval           data collection interval, default 5 sec
        --limit              collect given number of updates
        --normalized         normalize values to 100% on multi-CPU systems
        description: top shows a breakdown of the executables that triggered
                     Endpoint CPU usage within the last interval, multiple
                     instances are grouped together
        column abbreviations:
            MLWR       Malware Protection
            NET        Network Events
            PROC       Process Events
            FILE       File Events
            REG        Registry Events
            DNS        DNS Events
            LIB        Library Load Events
            AUTH       Authentication Events
            CRED       Credential Access Events
            RANSOM     Ransomware Protection
[...]

PR #5778, opened 2024-09-04
Adds Elastic Endpoint command reference

natasha-moore-elastic:
Resolves https://github.com/elastic/security-docs/issues/5097 by adding a new page that documents the commands for managing and troubleshooting Elastic Endpoint.

### Previews
* ESS: [Elastic Endpoint command reference](https://security-docs_bk_5778.docs-preview.app.elstc.co/guide/en/security/master/endpoint-command-ref.html)
* Serverless: Open the link in [this comment](https://github.com/elastic/security-docs/pull/5778#issuecomment-2331314083) and navigate to Security -> View serverless docs -> Manage Elastic Defend -> Elastic Endpoint command reference

### DIFF
--- a/docs/management/admin/endpoint-command-ref.asciidoc
+++ b/docs/management/admin/endpoint-command-ref.asciidoc
@@ -5,12 +5,12 @@ This page lists the commands for management and troubleshooting of {elastic-endp
 
 [NOTE]
 ====
-* The service is not added to the `PATH` system variable, so you must prepend the commands with the full OS-dependent path:
+* {elastic-endpoint} is not added to the `PATH` system variable, so you must prepend the commands with the full OS-dependent path:
 ** On Windows: `"C:\Program Files\Elastic\Endpoint\elastic-endpoint.exe"`
 ** On macOS: `/Library/Elastic/Endpoint/elastic-endpoint`
 ** On Linux: `/opt/Elastic/Endpoint/elastic-endpoint`
 
-* You must run the commands with elevated privileges—as the root user on POSIX systems, or as Administrator on Windows.
+* You must run the commands with elevated privileges—as the root user on Linux and macOS, or as Administrator on Windows.
 ====
 
 The following {elastic-endpoint} commands are available:
@@ -232,7 +232,7 @@ Fleet server: https://fleet.example.elastic.co:443
 [[elastic-endpoint-top-command]]
 == elastic-endpoint top
 
-Show a breakdown of the executables that triggered {elastic-endpoint} CPU usage within the last interval. This utility displays which {elastic-endpoint} features are resource-intensive for a particular executable.
+Show a breakdown of the executables that triggered {elastic-endpoint} CPU usage within the last interval. This displays which {elastic-endpoint} features are resource-intensive for a particular executable.
 
 NOTE: The meaning and output of this command are similar, but not identical, to the POSIX `top` command. The `elastic-endpoint top` command aggregates multiple processes by executable. The utilization values aren't measured by the OS scheduler but by a wall clock in user mode. The output helps identify outliers causing excessive CPU utilization, allowing you to fine-tune the {elastic-defend} policy and exception lists in your deployment.
 
--- a/docs/management/admin/endpoint-command-ref.asciidoc
+++ b/docs/management/admin/endpoint-command-ref.asciidoc
@@ -48,16 +48,40 @@ Gather diagnostics information from {elastic-endpoint}. This command produces an
 - `logs` directory: Copy of {elastic-endpoint} log files
 
 [discrete]
+=== Example
+
+[source,shell]
+------
+sudo /Library/Elastic/Endpoint/elastic-endpoint diagnostics
+------
+
+[discrete]
 [[elastic-endpoint-help-command]]
 == elastic-endpoint help
 
 Show help for the available commands.
 
 [discrete]
+=== Example
+
+[source,shell]
+------
+sudo /Library/Elastic/Endpoint/elastic-endpoint help
+------
+
+[discrete]
 [[elastic-endpoint-inspect-command]]
 == elastic-endpoint inspect
 
 Show the current {elastic-endpoint} configuration.
+
+[discrete]
+=== Example
+
+[source,shell]
+------
+sudo /Library/Elastic/Endpoint/elastic-endpoint inspect
+------
 
 [discrete]
 [[elastic-endpoint-install-command]]
@@ -77,6 +101,14 @@ Specify a resources `.zip` file to be used during the installation.
 Upgrade the existing installation.
 
 [discrete]
+=== Example
+
+[source,shell]
+------
+sudo /Library/Elastic/Endpoint/elastic-endpoint install --upgrade
+------
+
+[discrete]
 [[elastic-endpoint-memorydump-command]]
 == elastic-endpoint memorydump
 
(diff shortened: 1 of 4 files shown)


elasticdocs:
:rocket: Built _elastic-dot-co-docs-preview-docs_ successfully!

* https://elastic-dot-co-docs-production-ig5vpt8l7-elastic-dev.vercel.app

* Built with commit 9ae6d16d7a3ba5127106f00b6fb507354680ba61

> Issues? Visit #next-docs in Slack

docs/management/admin/endpoint-command-ref.asciidoc
```diff
@@ -0,0 +1,343 @@
...
+------
+sudo /Library/Elastic/Endpoint/elastic-endpoint test output
+------
+
+[discrete]
+=== Example output
+
+[source,txt]
+----
+Testing output connections using config file: [C:\Program Files\Elastic\Endpoint\elastic-endpoint.yaml]
```
intxgo, line 217:
```suggestion
Testing output connections
```
The path was appended up to 8.14, since 8.15 we're no longer using `yaml` config file so we changed the message.

intxgo (approved):

docs/management/admin/endpoint-command-ref.asciidoc
```diff
@@ -0,0 +1,343 @@
+[[endpoint-command-ref]]
+= {elastic-endpoint} command reference
+
+This page lists the commands for management and troubleshooting of {elastic-endpoint}, the installed component that performs {elastic-defend}'s threat monitoring and prevention.
+
+[NOTE]
+====
+* The service is not added to the `PATH` system variable, so you must prepend the commands with the full OS-dependent path:
```
joepeeples, line 8: The previous sentence calls it a "component" and this calls it a "service." It's not majorly confusing, but just feels inconsistent. We could just call it by name here:
```suggestion
* {elastic-endpoint} is not added to the `PATH` system variable, so you must prepend the commands with the full OS-dependent path:
```

docs/management/admin/endpoint-command-ref.asciidoc
```diff
@@ -0,0 +1,343 @@
...
+** On Windows: `"C:\Program Files\Elastic\Endpoint\elastic-endpoint.exe"`
+** On macOS: `/Library/Elastic/Endpoint/elastic-endpoint`
+** On Linux: `/opt/Elastic/Endpoint/elastic-endpoint`
+
+* You must run the commands with elevated privileges—as the root user on POSIX systems, or as Administrator on Windows.
```
joepeeples, line 13: Also unnecessarily technical, could be simpler:
```suggestion
* You must run the commands with elevated privileges—as the root user on Linux and macOS, or as Administrator on Windows.
```

docs/management/admin/endpoint-command-ref.asciidoc
```diff
@@ -0,0 +1,343 @@
...
+- `analysis.txt`: Diagnostic analysis report
+- `logs` directory: Copy of {elastic-endpoint} log files
+
+[discrete]
+=== Example
+
+[source,shell]
+------
+sudo /Library/Elastic/Endpoint/elastic-endpoint diagnostics
+------
```
joepeeples, line 56: The examples only include Linux/macOS. I assume admin-level readers wouldn't struggle with this, but it still feels incomplete to omit Windows.

For the serverless/MDX version, it's pretty easy to create a tabbed widget that includes Windows too ([example](https://www.elastic.co/docs/current/serverless/security/uninstall-agent), [code](https://github.com/elastic/security-docs/blob/main/docs/serverless/edr-install-config/uninstall-agent.mdx?plain=1#L17-L28)), but tabs are kind of a nightmare for AsciiDoc. Maybe we could get away with omitting Windows just for AsciiDoc, or use a simpler soft heading format ([example](https://www.elastic.co/guide/en/security/master/uninstall-agent.html#uninstall-endpoint)).

docs/management/admin/endpoint-command-ref.asciidoc
```diff
@@ -0,0 +1,343 @@
...
+sudo /Library/Elastic/Endpoint/elastic-endpoint inspect
+------
+
+[discrete]
+[[elastic-endpoint-install-command]]
+== elastic-endpoint install
+
+Install {elastic-endpoint} as a system service.
+
+NOTE: Elastic doesn't publish independent {elastic-endpoint} packages since {elastic-endpoint} is managed by {agent}.
```
joepeeples, line 92: What exactly should the user conclude from this, and what action do we want them to take with this information? Should they _not_ install Endpoint like this, if it's limited in this way? That's kinda what it sounds like, but it's so indirect it's hard to tell.

docs/management/admin/endpoint-command-ref.asciidoc
```diff
@@ -0,0 +1,343 @@
...
+
+Fleet server: https://fleet.example.elastic.co:443
+        Status: Success
+----
+
+[discrete]
+[[elastic-endpoint-top-command]]
+== elastic-endpoint top
+
+Show a breakdown of the executables that triggered {elastic-endpoint} CPU usage within the last interval. This utility displays which {elastic-endpoint} features are resource-intensive for a particular executable.
```
joepeeples, line 235: Another word choice nit: this is just a single command, but typically I think of a "utility" more as a small application that performs several different actions or commands.
```suggestion
Show a breakdown of the executables that triggered {elastic-endpoint} CPU usage within the last interval. This displays which {elastic-endpoint} features are resource-intensive for a particular executable.
```

joepeeples (commented):
Great chunk of content that I'm sure customers and SMEs will appreciate! Added a few comments to the AsciiDoc version, which also apply to MDX.

docs/management/admin/endpoint-command-ref.asciidoc
intxgo: Hi, I wrote the draft, the point here was that users reading this documentation shouldn't start googling (or asking in the community) where to get the Endpoint installation package from. Indeed it also implies they have no business trying to install it by this command. Endpoint is unusable without Agent.

I wont mind to omit this command from the docs entirely if it makes more sense 😉

docs/management/admin/endpoint-command-ref.asciidoc
intxgo: Can't we just drop the prefix? The note at the top of the article was explaining this, so users should figure it out easily what to do when this won't work for them
```
elastic-endpoint diagnostics
```

joepeeples: > Can't we just drop the prefix? 

Sounds good! Keeps things simple and clear. Maybe update the note to remind readers they'll need `sudo` to run as the root user? Again, the audience is likely tech-savvy enough to know this, but wouldn't hurt to remind them too.

docs/management/admin/endpoint-command-ref.asciidoc
joepeeples: Thanks for the helpful context @intxgo! I could see us either omitting the command for those reasons, or including it but with an admonition that basically tells the user don't bother using it because standalone Endpoint (without Agent) won't work. 

The main reason I suggest including it is for completeness; if the user can run `elastic-endpoint help` and the output includes `install`, then they might be tempted to try it out.

intxgo: agree, just feel free to reword the note part to make it clearer

joepeeples (approved):
:shipit:

PR #5826, opened 2024-09-13
Remove Development Rules from 8.13 security docs

shashank-elastic:
- As part of https://github.com/elastic/detection-rules/issues/4008 we plan to remove all prebuilt rule references for development rules. 
- This PR targets the 8.13 documents.


approksiu (approved):
LGTM!

PR #5834, opened 2024-09-16
Update .mergify.yml for `8.16`

jmikell821:
Adds mergify option for 8.16, which is to branch 8x.


joepeeples:
Do we need a similar new entry to add `8.x` as well? Even if 8.x will only exist for a short time, we'll probably want to use Mergify to keep it up to date, same as 8.16 and other version branches.

lcawl (approved):
LGTM, matches what we put in stack-docs

joepeeples (approved):

PR #5959, opened 2024-10-18
[8.16] Update .mergify.yml

jmikell821:
Updating mergify for `8.16`.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -23,7 +23,7 @@ pull_request_rules:
         assignees:
           - "{{ author }}"
         branches:
-          - "8.x"
+          - "8.16"
         title: "[{{ destination_branch }}] {{ title }} (backport #{{ number }})"
         labels:
           - backport
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -13,6 +13,20 @@ pull_request_rules:
             git merge upstream/{{base}}
             git push upstream {{head}}
             ```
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -13,6 +13,20 @@ pull_request_rules:
...
+    conditions:
+      - merged
+      - base=main
+      - label=v8.16.0
+    actions:
+      backport:
+        assignees:
+          - "{{ author }}"
+        branches:
+          - "8.x"
```
nastasha-solomon, line 26: Since `8.x` is now `8.17`, shouldn't this be `8.16`?

```suggestion
          - "8.16"
```

nastasha-solomon: Actually, scratch this...the current value (`8.x`) might be correct since we haven't cut an `8.17` branch from `8.x` yet. I  saw https://github.com/elastic/security-docs/pull/5958 and automatically jumped to the thought that the `8.x` branch is now synonymous with `8.17`.

lcawl: I think so too, though it's maybe moot since this PR is against the 8.16 branch so it won't be backporting to itself :)

nastasha-solomon: Yeah, that's a good point, @lcawl. I would still advise that we change `8.x` to `8.16` though, so it's clear that 8.16 is the target branch for this action.

nastasha-solomon (approved):

PR #6180, opened 2024-11-19
Update serverless Discover URL

leemthompo:
👋 We're de-duplicating content in the Elasticsearch serverless docs, and this is an example of a feature that will have its single-source-of-truth in the Kibana docs. I'll need this change merged, in order to get a green build on my PR in the `docs-content` repo.


natasha-moore-elastic (approved):
LGTM!

PR #6219, opened 2024-11-21
[8.x] Update .backportrc.json

jmikell821:
Adds `8.17` to list of backports.


lcawl (approved):
LGTM

nastasha-solomon (approved):
Approving this since @lcawl gave the thumbs up. But to be transparent, I don't fully understand why we're updating this file in the `8.x` branch and how it'll affect PRs that are merged directly into `8.x` instead of `main`.

lcawl:
> But to be transparent, I don't fully understand why we're updating this file in the `8.x` branch and how it'll affect PRs that are merged directly into `8.x` instead of `main`.

Per https://github.com/sorenlouv/backport/blob/main/docs/config-file-options.md#project-config-backportrcjson that section of the file is just the "List of target branches the user can select interactively".  Thus if someone has the 8.x branch checked out and runs the `backport` command, it'll now include v8.17 in the list of possible target branches.

nastasha-solomon:
Oh nice, I didn't know that file existed and completely forgot about our internal backport tool (we've mainly workign with Mergify 😅).  Thanks for sharing the docs @lcawl. The updates @jmikell821 made make much more sense now.

Issue #6365, opened 2024-12-18
[Request] Crowdstrike additional third-party response actions


raqueltabuyo:
### What can we change to make the docs better?

**Description**
We are adding new third-party actions to Crowdstrike response actions, which will allow users to execute remote commands using Crowdstrike agent through Elastic Security.

This is similar to the functionality (and docs) we previously added for Sentinel One and Crowdstrike: https://www.elastic.co/guide/en/security/current/third-party-actions.html

**Background & resources**
PRs:
- https://github.com/elastic/kibana/pull/201420
- https://github.com/elastic/kibana/pull/202012
- https://github.com/elastic/kibana/pull/203101
- https://github.com/elastic/kibana/pull/203420
- https://github.com/elastic/kibana/pull/204044

Issues/metas: https://github.com/elastic/security-team/issues/10873
Point of contact: @caitlinbetz @tomsonpl @raqueltabuyo @ashokaditya @paul-tavares
Test environments:

### Doc URL

This is similar to the functionality (and docs) we previously added for Sentinel One and Crowdstrike: https://www.elastic.co/guide/en/security/current/third-party-actions.html
Github issue link(s)/Other resources:
 https://github.com/elastic/security-team/issues/10873

### Which documentation set needs improvement?

ESS and serverless

### Software version

ESS release
8.18

Serverless release
January 27, 2025

Feature differences
Feature will be the same in serverless/ESS

ESS release: 8.18

API docs impact
TBD

Prerequisites, privileges, feature flags
ESS & Serverless, Kibana privileges:

Security solution privilege: TBD

Actions and Connectors privilege:: EDR Connectors

tomsonpl:
@natasha-moore-elastic Hey 👋 
I tried to leave as much details as possible in the PRs' description, but please reach out if there's something you'd like to talk about :)

natasha-moore-elastic:
Hi @tomsonpl! I have a few questions to start with:

* Is there an environment you could share that we can use for testing out this feature?
* This https://github.com/elastic/kibana/pull/202012 seems to be the main PR that introduces the `runscript` response action command. Are there any other new commands that we need to document?
* I’m not familiar with the concept of sub actions (mentioned in https://github.com/elastic/kibana/pull/203420), and I don’t see any sub actions documented in our current docs. Is this a new concept and do we need to start documenting those?

tomsonpl:
Hey 👋 
[...]

PR #6435, opened 2025-01-17
Adds new runscript Crowdstrike response action

natasha-moore-elastic:
Resolves https://github.com/elastic/security-docs/issues/6365.

Previews

### DIFF
--- a/docs/management/admin/response-actions.asciidoc
+++ b/docs/management/admin/response-actions.asciidoc
@@ -193,6 +193,25 @@ Example: `scan --path "/Users/username/Downloads" --comment "Scan Downloads fold
 NOTE: Scanning can take longer for directories containing a lot of files.
 
 [discrete]
+[[runscript]]
+=== `runscript`
+
+Run a script on a host. You must include one of the following parameters to identify the script you want to run:
+
+* `--Raw`: The full script content provided directly as a string.
+* `--CloudFile`: The name of the script stored in a cloud storage location.
+* `--HostPath`: The absolute or relative file path of the script located on the host machine.
+
+You can also use these optional parameters:
+
+* `--CommandLine`: Additional command-line arguments passed to the script to customize its execution.
+* `--Timeout`: The maximum duration, in seconds, that the script can run before it's forcibly stopped. If no timeout is specified, it defaults to TBD.
+
+Required privilege: TBD
+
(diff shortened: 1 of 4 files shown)


raqueltabuyo (approved):

tomsonpl (commented):
Hey @natasha-moore-elastic, this looks awesome ❤️ 
Leaving some thoughts here:
1. Runscript response action - currently it is on the list of all response actions, however it's the first response action that is not supported by Endpoint, but only on CrowdStrike - should we distinguish this somehow, or at least mention 'CrowdStrike only' or something like that ?
2. In SentinelOne, we have a information about: 
```View past response action activity in the [response actions history](https://security-docs_bk_6435.docs-preview.app.elstc.co/guide/en/serverless/main/security-response-actions-history.html) log.```
We are capable of doing this for CrowdStrike too, so we could think of adding it to the docs too, what do you think? 
3. runscript examples - can we add 3 examples ? 1 per each argument : Raw, HostPath, CloudFile ? 
4. I was wrong when confirming the predefined roles for RunScript. They should not be the same as for Isolate, but the same as for Execute - meaning: `Predefined role: SOC manager or Endpoint operations analyst`

File: docs/management/admin/response-actions.asciidoc [hunk 1]
```diff
@@ -209,7 +211,13 @@ You can also use these optional parameters:
 
 Required privilege: **Execute Operations**
 
-Example: `runscript --CloudFile="CloudScript1.ps1" --CommandLine="-Verbose true" --Timeout=180` 
+Examples:
+
+`runscript --CloudFile="CloudScript1.ps1" --CommandLine="-Verbose true" --Timeout=180`
+
+`runscript --Raw="Get-ChildItem."`
```
tomsonpl, line 218: best to wrap the string in triple backticks ```Get-ChildItem.```, it's the syntax CS prefers.

tomsonpl (dismissed):
Thanks for the changes 👍 I left one minor thing, otherwise LGTM :) This is fantastic ❤️

docs/management/admin/response-actions.asciidoc
tomsonpl: but only for --Raw type

nastasha-solomon (approved):
lgtm!

PR #6466, opened 2025-01-22
Fix doc indentation 8.14

shashank-elastic:
As part of the PR - https://github.com/elastic/security-docs/pull/6447 there was some indetation issues like below 
image

This PR fixes that issue


shashank-elastic:
Preview From the Fix is working as expected 

image

Aegrah (approved):

PR #6468, opened 2025-01-22
Fix doc indentation 8.16

shashank-elastic:
As part of the PR https://github.com/elastic/security-docs/pull/6452 there was some indentation issues like below 

image

This PR fixes that issue


shashank-elastic:
Preview of the fix is working as expected 

image

PR #6507, opened 2025-02-05
Adds `8.18` backport option to mergify

jmikell821:
Adds `8.18` backport option to mergify.


.mergify.yml
```diff
@@ -13,6 +13,20 @@ pull_request_rules:
             git merge upstream/{{base}}
             git push upstream {{head}}
             ```
+  - name: backport patches to 8.18 branch
+    conditions:
+      - merged
+      - base=main
```
nastasha-solomon, line 19: This, I'm not 100% sure about since we'll no longer have a `main` branch to create PRs from. @lcawl should this be `8.x` instead? If so, do we need to update all other instances in this file that have the `base` branch set to `main`?

nastasha-solomon: Nvm, this is probably not an issue!

lcawl: I remember a slack discussion that said this `base` clause was removed from the file in the `main` branch since mergify was only looking at that branch... which I found to be a very confusing discussion.

But i guess worst case we keep our eyes open for the first PR that starts in 8.x and see if it's successfully backported. If not, we might need to remove or change this `base` clause.

PR #6558, opened 2025-02-28
Fix ES date time ref in prebuilt rule docs

nastasha-solomon:
TBD


nastasha-solomon:
No longer needed as files have moved.

PR #6559, opened 2025-03-01
Add applies_to frontmatter

lcawl:
This PR is intended to show the type of [applies_to](https://docs-v3-preview.elastic.dev/elastic/docs-builder/tree/main/syntax/applies) frontmatter that should be included in security reference content (in particular, so that it can be considered in updates to the automated content).

I noticed that there seemed to be at least one deprecated rule, so I updated the frontmatter there appropriately:

```
---
applies_to:
  stack: deprecated 8.14.0
  serverless:
    security: deprecated all
---
```
... which is currently displayed as follows:

image

I've used `all` for now but going forward per https://docs-v3-preview.elastic.dev/elastic/docs-builder/tree/main/versions/#versions-and-lifecycle-states the appropriate version labels should be used when a feature is added. For example, if there's a new rule added in 9.0.0 or later, that V.R.M should be used in the frontmatter.


lcawl:
Closing since the files are now moved to a new repo and frontmatter will need to be automated.

PR #6680, opened 2025-03-28
Deletes Endpoint management API

natasha-moore-elastic:
Contributes to https://github.com/elastic/security-docs-internal/issues/59.
Deletes Endpoint management API docs from 8.16+ branches.


nastasha-solomon (approved):
thanks for cleaning these up!

PR #6911, opened 2025-07-08
Removes statement about CSPM only working in default space

benironside:
Minor fix as part of docs-content/2036


jmikell821 (approved):
LGTM!

PR #7142, opened 2026-02-18
[Security][Cases][Bug]: Case templates were GA'd in 8.18

nastasha-solomon:
Fixes https://github.com/elastic/docs-content/issues/5209. 

Case templates were moved from tech preview to GA in 8.18. The 8.18 and 8.19 docs still show them as being in tech preview.


benironside (approved):
LGTM

